PR #57, opened 2020-01-28
Add support for configuring where the front-end finds the back-end.

havfo:
The back-end location can be configured in the config.js file.


havfo:
Fixed. I also put config.js into .gitignore because you wouldn't normally want to check in that file.

PR #75, opened 2020-06-19
Statically typed API client and some small fixes

jorgenbele:


src/auth.tsx
```diff
@@ -1,22 +1,33 @@
...
+    if (callback) callback();
   }
-  logout(cb: any) {
-    localStorage.clear();
-    cookies.remove('token'); 
+  logout(callback?: () => void) {
     this.authenticated = false;
-    cb();
+    this._token = undefined;
+    cookies.remove("token");
```
ddabble, line 22: The logout endpoint in the backend should also be called.

ddabble: Moving this change to #79 instead of this PR.

PR #90, opened 2020-07-01
Improvements in Notification Profile view and close issues,.

jorgenbele:
1. Change defaults for new profiles
2. Fix problems when connection to backend is down/action fails.
3. Fix problem where available timeslots doesn't update correctly when a profile is deleted. 
4. Update index.html and manifest.json to display Argus instead of "React App" or equivalent.


src/components/profilelist/ProfileList.tsx
```diff
@@ -257,12 +282,13 @@ const ProfileList: React.FC = () => {
...
               return (
                 <Profile
+                  unsavedChanges={unsavedProfiles.has(profile.timeslot.pk)}
                   exists
                   active={profile.active}
                   filters={filters || new Map<FilterPK, Filter>()}
                   timeslots={timeslots || new Map<TimeslotPK, Timeslot>()}
                   isTimeslotInUse={(timeslot: Timeslot): boolean => usedTimeslots.has(timeslot.name)}
-                  key={profile.pk}
+                  key={JSON.stringify(profile)}
```
jorgenbele, line 291: This is not the best way to do things, but it works and is simple. Shouldn't be a problem unless lots of profiles are in use.

jorgenbele: This has been replaced, so it's not relevant anymore.

jorgenbele:
I'm just merging it now. Any discussions on what the title and such should be can be done in the original issues/new issues.

PR #110, opened 2020-08-27
Display events, acknowledgements and tags & proper filtering based on tags

jorgenbele:
Incident details modal:
 - Fetches acknowledgements and other  related events from Argus API

 - Displays the acks and events in a nice list

 - The "Create acknowledgement" button now will actually create an acknowledgement.

Tag filtering:
 - Tags of the same "type" (same key) works
   as an OR, meaning that if the following
   tags are used:

     host=testhost1, host=testhost2

   Incidents with either the tag "host=testhost1"
   or the tag "host=testhost2" will be shown,
   but all other values for host will be filtered
   out.

 - Tags of different "types" works as an AND,
   meaning that if the following tags are used:

     host=testhost1, host=testhost2,
     ticket_url=http://tr.t/123

   Only the incidents with host in (testhost1, testhost2)
   AND with ticket_url equal to http://tr.t/123
   will be displayed.

### DIFF
--- a/src/api/hooks.tsx
+++ b/src/api/hooks.tsx
@@ -3,11 +3,13 @@ import { Incident } from ".";
 // TODO: fix this
 import { incidentWithFormattedTimestamp, IncidentWithFormattedTimestamp } from "../utils";
 import {
+  Acknowledgement,
   ApiErrorType,
-  NotificationProfile,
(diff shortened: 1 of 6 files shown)


havfo (approved):
Looks good!

PR #111, opened 2020-08-28
Implement "show acked incidents" checkbox for incident view

jorgenbele:
This now makes it possible to conditionally display the acknowledged incidents, but will by default only display open+unacknowledged incidents.

When the "open" option for "show open/closed" is used it will use the "open+unacked" endpoint when this checkbox option is disabled. When it is enabled it will use the "open" endpoint.

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -475,6 +475,14 @@ export class ApiClient {
     );
   }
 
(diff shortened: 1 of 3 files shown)


hmpf (approved):
Tested by hand, seems to work.

PR #112, opened 2020-09-03
Phone number support

havfo:


hmpf (commented):
The spinner in the create-button doesn't stop spinning after a new phone number is successfully added. Reloading the page stops it and greys out the button.

hmpf (approved):
Tested manually, works.

PR #113, opened 2020-09-10
Remove all traces of object, parentObject, problemType

hmpf:


jorgenbele (approved):
These changes together remove the old object, parentObject, problemType based filters with the new tag based system. I replaced the table component with a MUI table component so that it matches the one used for incidents, and changed to using the incidents endpoint with query parameters instead of a POST request to previewFilter.

PR #114, opened 2020-09-10
Make the first day of the week be number 1

hmpf:
See https://github.com/Uninett/Argus/issues/140


jorgenbele (approved):
Seems good! Tested with success locally.

PR #123, opened 2020-09-16
Update Event type and display username in events instead of pk

jorgenbele:


hmpf (approved):
Manually tested, does what it says on the tin.

PR #132, opened 2020-09-18
Stop showing empty ticket url as error

### LINKED ISSUE #118: Polish: ticket url in details-box
1. When opening the details box, the ticket url is highlighted in red and marked as invalid. Not having a ticket-url is perfectly valid!
2. Clicking on the red line moves "ticket" up and down.
3. It looks like it is possible to edit the ticket without clicking "Edit" first. Fixing 1 might also fix this.

hmpf:
Closes #118

### DIFF
--- a/src/components/incident/IncidentDetails.tsx
+++ b/src/components/incident/IncidentDetails.tsx
@@ -155,7 +155,7 @@ const TicketModifiableField: React.FC<TicketModifiableFieldPropsType> = ({
     setChangeUrl(false);
   };
 
-  const error = useMemo(() => !isValidUrl(url || ""), [url]);
+  const error = useMemo(() => !(url || !isValidUrl(url || "")), [url]);
 
   return (
     <ListItem>


lunkwill42 (commented):
I can't say whether this actually works, but I approve of the sentiment :-)

jorgenbele (changes requested):
Same thing here as in other PRs. I don't think isValidUrl should have an emptyOk parameter. Instead it should be used like this `(!!url && !isValidUrl(url || ""))`  which evaluates to false when url is empty/null/undefined.

hmpf:
> Same thing here as in other PRs. I don't think isValidUrl should have an emptyOk parameter. Instead it should be used like this `(!!url && !isValidUrl(url || ""))` which evaluates to false when url is empty/null/undefined.

Aha, I guess there's not enough context here to see why it can't be like that here at a glance. An empty _ticket url_ is **valid** and should not trigger an error. A non empty absolute _ticket url_ is also **valid**. What is **not valid** is a _non-empty string that is not an url_.

jorgenbele:
> > Same thing here as in other PRs. I don't think isValidUrl should have an emptyOk parameter. Instead it should be used like this `(!!url && !isValidUrl(url || ""))` which evaluates to false when url is empty/null/undefined.
> 
> Aha, I guess there's not enough context here to see why it can't be like that here at a glance. An empty _ticket url_ is **valid** and should not trigger an error. A non empty absolute _ticket url_ is also **valid**. What is **not valid** is a _non-empty string that is not an url_.

The non-empty string that is not a valid URL:
` !!"test" && !false == true` 

The empty-string case:
` !!"" == false`

The non-empty string that is a valid URL:
` !!"some-absolute-url" && !true == false`

As you can see only the case where there the expression evaluates to true is when the string is non-empty and NOT a URL.
Is this not the correct behaviour that you describe? `error = <expr>` means that it's only an error in the scenarios where the expression evaluates to true.

hmpf:
"!!"  was new to me. Modern Javascript rookie, remember? Now that was fun to google, standard way to cast to boolean. Yay. Time to test.

For my future self: If the url is (non-empty and non-valid) `x && !isValidUrl(x)` we raise an error. If it is (empty or valid) `!x || isValidUrl(x)` do not raise an error. In the first case, if `x` is the empty string we get the empty string, not a booelan, so cast it to a boolean by `!!`, so `!!x && !isValidUrl(x)`. This is a _nightmare_ to read. (I cannot seem to google up the article that greatly influenced my thought on how to do booelan expressions. It went something like this: negation is confusing so rewrite to keep the number of NOTs down.)  Either way, we don't need to pass in an empty string to `isValidUrl` if my js-fiddling is to be believed, the API always sends a string.

De Morgan to the rescue: `!(x && !isValidUrl(x))`. Which is the most readable?




 `!!url && !isValidUrl(url`, `!(x && !isValidUrl(x))` and `Boolean(x) && !isValidUrl(x)`

hmpf:
I picked 2), with the guard, so: `!(x || !isValidUrl(x || ""))`

PR #135, opened 2020-09-18
Improve title of incident detail box

### LINKED ISSUE #121: Polish: the header/title of the incident detail box is useless
The id of an incident internally in argus is utterly irrelevant for a human end user.

Better to use the summary/description.

hmpf:
Closes #121 

It's a little odd that this part of the details box is defined in the incident table, not the incident detail box itself. Not something we should bother to fix before 1.0 though.

@lunkwill42, I've tagged you because exactly what we ought to have as the title of the box is very much up for discussion. It's very easy to change though, as long as only things stored directly on the incident are used.

### DIFF
--- a/src/components/incidenttable/IncidentTable.tsx
+++ b/src/components/incidenttable/IncidentTable.tsx
@@ -25,7 +25,7 @@ import TableHead from "@material-ui/core/TableHead";
 import TableRow from "@material-ui/core/TableRow";
 import TablePagination from "@material-ui/core/TablePagination";
 
-import { useStateWithDynamicDefault, toMap, pkGetter } from "../../utils";
+import { useStateWithDynamicDefault, toMap, pkGetter, truncateMultilineString } from "../../utils";
 
 import { useAlertSnackbar, UseAlertSnackbarResultType } from "../../components/alertsnackbar";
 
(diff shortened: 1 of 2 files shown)


jorgenbele:
The reason that it's defined in the view instead of the details component is so that the details component can be reused wherever it's useful to inline details about an incident. It is part of the dialog/modal title, not the details component, because it might not be useful to display a title and/or want it to be a dialog/modal.

jorgenbele (approved):

lunkwill42 (changes requested):
I agree that using the Incident description as the dialog box title is much better, **however**:

1. We have no limits to how long a description may be
2. A description may even be multi-line

So to use this as title, it would need to be filtered. Suggestions:

1. If the description is multi-line, use only the first line.
2. If the description is longer than XX characters (lets pick the arbitrary number 50?), the rest should be truncated and replaced with an ellipsis...

hmpf:
Titles are now truncated, and the internal argus id is completely gone. @jorgenbele please check that the function I made is ok.

Prior to this was an octopus-merge that clobbered each other, so I fixed that too straight on master, in "Fix missing "}"". NTS: Avoid octopus-merges even more.

hmpf:
I hard-coded the length to 50 btw. where the function is used. If that constant should be somewhere else, feel free to move it @jorgenbele, because I don't know where it would belong.

lunkwill42 (approved):

PR #141, opened 2020-09-22
Polish: Remove mentions of Slack

### LINKED ISSUE #140: Polish: Remove any mention of slack
We're not supporting Slack yet, so remove it everywhere.

Problem: ProfileList.mediaOptions is hard coded. For media where everything can be configured simply (like for email or sms since there is table support for these) these should be looked up via an endpoint. If the configuration of media is made very dynamic and possible via this front-end, a new front-end must be released for every new notification media. Something to keep in mind for https://github.com/Uninett/Argus/issues/121.

hmpf:
Closes #140.

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -81,7 +81,7 @@ export const EmptyFilterDefinition = {
   sourceIds: [],
 };
 
-export type MediaAlternative = "EM" | "SM" | "SL";
+export type MediaAlternative = "EM" | "SM";
 
 export type PhoneNumberPK = number;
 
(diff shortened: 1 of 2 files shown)


jorgenbele (changes requested):
This requires that the MediaAlternative type is changed as well: https://github.com/Uninett/Argus-frontend/blob/3e6306f88fdea0ee7e4355d15ee0f1b8153bdf2f/src/api/index.tsx#L84. Remove "SL" from here.

hmpf:
I'm so making an endpoint in the API for this. Media need to be configurable at least from the admin. *hmpf*

jorgenbele (approved):

PR #144, opened 2020-09-25
Only linkify details_url if absolute

### LINKED ISSUE #125: Hyperlink to source system incident details is broken
As mentioned in https://github.com/Uninett/Argus-frontend/issues/93#issuecomment-692638392, the hyperlink to the source system's incident details on the incident details model is broken.

The Incident record itself may contain relative URLs. If a URL is not a fully qualified URL, the hyperlink should be constructed using the `base_url` present in the source system definition as a prefix. Example output from the Incident endpoint:

```json
[
  {
    "pk": 18,
    "start_time": "2020-09-14T17:06:49.983721+02:00",
    "end_time": "infinity",
    "source": {
      "pk": 1,
      "name": "testnav",
      "type": {
        "name": "nav"
      },
      "user": 2,
      "base_url": "http://localhost:9080/"
    },
    "source_incident_id": "199420",
    "details_url": "/search/event/199420",
    "description": "Link DOWN on A1 at example-sw.example.org (4.kx042)",
    "ticket_url": "",
    "tags": []
  }
]
```

The URL **may** also be a fully qualified URL, in which case it is ok to use it unchanged.

### LINKED ISSUE #143: Do not linkify relative details_urls
Closes #125.

Once https://github.com/Uninett/Argus/pull/161 is merged, the filterable incident endpoints will send three different types of `details_url`s:

* Empty string
* Relative url, because the source system does not have base_url set, and the source sends relative urls
* Absolute url

Relative urls should not be turned into links.

hmpf:
Closes #143, independently of what the backend does.

Closes #125 as well.

An empty details_url field is shown by an en dash in the UI.

### DIFF
--- a/src/components/incident/IncidentDetails.tsx
+++ b/src/components/incident/IncidentDetails.tsx
@@ -106,6 +106,14 @@ const isValidUrl = (url: string) => {
   return true;
 };
 
+const formatUrlOrIdentity = (url: string, title: string, empty: string = "") => {
+  if (isValidUrl(url || "")) {
+    return <a href={url}>{title}</a>;
+  } else {
+    return url || empty;
+  }
+};
+
 const TagChip: React.FC<TagChipPropsType> = ({ tag, small }: TagChipPropsType) => {
   if (isValidUrl(tag.value)) {
     return (
@@ -450,7 +458,7 @@ const IncidentDetails: React.FC<IncidentDetailsPropsType> = ({
                   <IncidentDetailsListItem title="Source" detail={incident.source.name} />
                   <IncidentDetailsListItem
                     title="Details URL"
-                    detail={<a href={incident.details_url}>{incident.details_url}</a>}
+                    detail={formatUrlOrIdentity(incident.details_url, "More details", "–")}
                   />
 
                   <TicketModifiableField
--- a/src/components/incident/IncidentDetails.tsx
+++ b/src/components/incident/IncidentDetails.tsx
@@ -106,11 +106,11 @@ const isValidUrl = (url: string) => {
   return true;
 };
 
-const formatUrlOrIdentity = (url: string, title: string, empty: string = "") => {
-  if (isValidUrl(url || "")) {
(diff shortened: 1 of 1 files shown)


src/components/incident/IncidentDetails.tsx
```diff
@@ -106,6 +106,14 @@ const isValidUrl = (url: string) => {
   return true;
 };
 
+const formatUrlOrIdentity = (url: string, title: string, empty: string = "") => {
+  if (isValidUrl(url || "")) {
+    return <a href={url}>{title}</a>;
+  } else {
+    return url || empty;
```
jorgenbele, line 113: Instead of returning `url || empty` here, you could just do 
`details={url && <a href={incident.details_url}>More details</a>}` if you want empty string on empty url string or 
`details={url && <a href={incident.details_url}>More details</a> || "-"}` if you want "-" when url is "" (empty string).
The function name "formatUrlorIdentity" is misleading, instead `hyperlinkIfUrl(url, title)` or something and instead of returning `url || empty` just use it like this `hyperlinkIfUrl(url, title) || "-"`.

jorgenbele (changes requested):
I think the function should be renamed to something more clear, and it makes more sense to "||" the result of the function call to get another value instead of passing another parameter and doing it inline.

src/components/incident/IncidentDetails.tsx
hmpf: Ugh, copypaste-bug `isValidUrl(url || "")` should be `isValidUrl(url)`. We only want a link if the url is absolute.

hmpf: I tried doing stuff right in the return but got so many errors it was easier to make a function.

hmpf: `isValidUrl(url) && <a href={incident.details_url}>More details</a> || "-"` maybe.

hmpf: I went with renaming and simplifying the function.

jorgenbele (approved):

PR #155, opened 2020-10-13
Feat new timeslots

### LINKED ISSUE #142: Frontend must support multiple days per timeslot
See https://github.com/Uninett/Argus/issues/140.

jorgenbele:
Closes #142


hmpf (approved):
The UX is very clunky, easy to get wrong. But, we need to get this out of the door, with use we might think up a better way of doing it. It's not that clear that it is necessary to add a recurrence.

Have you seen how it's done in the admin? Also not that good but maybe easier to see what needs to be done.

This does give me an idea however, a "library" of recurrences that people can choose from. ("Every day", "Weekend", "Working days", for starters.) But not for version 1.0!

hmpf:
Closes #142.

Closes https://github.com/Uninett/Argus/issues/140.

jorgenbele:
> It's not that clear that it is necessary to add a recurrence.
I agree. Maybe a title "Recurrences" above the list of recurrences would help. Displaying "No recurrences" when the list is empty might also be reasonable.

Converting the "Add recurrence" button to a Floating Action Button and moving it below the list of recurrences is another potential option. 

> Have you seen how it's done in the admin? 
I just tested how it's implemented in django admin, and I can see some advantages in it's use of checkboxes to select days. I will try to change the AutoComplete component to a multi-Select component. That way it's easier to select multiple days but it is still displayed in a compact way.

As for the "add timeslot" element I think a default recurrence is useful to make it more clear that a recurrence is required.

> This does give me an idea however, a "library" of recurrences that people can choose from
This sounds like a good idea and a pure frontend implementation with a library of hardcoded recurrences to choose from shouldn't be that hard to implement.

hmpf:
I'll merge this now so that we have *something*.

PR #160, opened 2020-10-21
Switch to Select component in timeslots

jorgenbele:
UX improvements by
- switching from AutoComplete component to Select component
- aligning subcomponents by using Grid's
- disable components on update


hmpf (approved):
Tested and approved

PR #179, opened 2020-12-15
Add a complete Docker Compose-based development environment

lunkwill42:
This add a full Docker Compose-based environment for Argus frontend development, which includes a full backend setup to work against.

Some tweaks and bugfixes are included, as they were blocking the completion of the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ Set up the Argus backend according to the instructions in its repository (https:
 
 Furthermore, [Node.js](http://nodejs.org/) is required. We also use the [Node Package Manager (npm)](https://www.npmjs.com/), which comes bundled with Node.js.
 
+**Optionally**, you can forego a full installation of Node.js and npm on your local system, and instead opt to run a complete Argus setup using [Docker Compose](https://docs.docker.com/compose/). If so, skip directly to the *Install and startup* section.
+
 Installation procedures are as follows:
 
 ### Install Node.js
@@ -77,6 +79,19 @@ This will open your browser at http://localhost:3000 or similar.
 Congrats, you now have the Argus frontend up and running!
 
 Note that the website will automatically reload as you edit the code.
+
+### Using Docker Compose
+
+This repository contains a `docker-compose.yml` definition to run all the backend components as services, while the Argus frontend runs directly off the checked out source code using npm. To accomplish this, all you need should be:
+
+    export UID
+	docker-compose up
+
+(the `export UID` step is to ensure the Argus frontend container runs using your system UID, so it doesn't produce root-owned files in the mounted source code directory).
+
+Your Argus frontend should now be served on `http://localhost:8080`, while your Argus API server should be served on `http://localhost:8000`. As with running `npm` locally, the website should automatically reload as you edit the code.
+
+The default setup will install the latest version of the Argus API server from the master branch. If you need to customize which tag or branch to install, you can change the `BRANCH` argument in `docker-compose.yml` (or preferably implement your own `docker-compose.override.yml`).
 
 ## Coding guidelines
 
--- a/README.md
+++ b/README.md
@@ -82,10 +82,10 @@ Note that the website will automatically reload as you edit the code.
 
 ### Using Docker Compose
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -78,6 +80,19 @@ Congrats, you now have the Argus frontend up and running!
 
 Note that the website will automatically reload as you edit the code.
 
+### Using Docker Compose
+
+This repository contains a `docker-compose.yml` definition to run all the backend components as services, while the Argus frontend runs directly off the checked out source code using npm. To accomplish this, all you need should be:
+
+    export UID
+	docker-compose up
```
hmpf, line 88: Is this something you type in the shell? If so, could you make it clearer? If not, what is this? "all you need should be:" how exactly?

docker/api/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# in the upstream repo
+FROM python:3.8
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y --no-install-recommends tini build-essential
+
+# Install argus from configured branch/tag and lock in some dev requirements
+# that aren't part of the regular setup.cfg:
+ARG BRANCH=master
+RUN pip install psycopg2-binary channels==2.4.0 django-extensions python-dotenv git+https://github.com/Uninett/argus@${BRANCH}
```
hmpf, line 11: Docker can use `.env`-files, yes. If you do, please document it.

docker/api/docker-entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -xe
+
+django-admin collectstatic --noinput
+django-admin migrate --noinput
+exec daphne -b 0.0.0.0 -p $PORT argus.ws.asgi:application
```
hmpf, line 5: How will I fill the empty backend with data when using docker-compose? Please document.

src/config.tsx
```diff
@@ -1,5 +1,5 @@
 export const BACKEND_URL = process.env.REACT_APP_BACKEND_URL || "";
-export const ENABLE_WEBSOCKETS_SUPPORT = false;
+export const ENABLE_WEBSOCKETS_SUPPORT = process.env.REACT_APP_ENABLE_WEBSOCKETS_SUPPORT === "true" || false;
 export const BACKEND_WS_URL = process.env.REACT_APP_BACKEND_WS_URL || "";
-export const USE_SECURE_COOKIE = process.env.REACT_APP_USE_SECURE_COOKIE === "false" || true;
+export const USE_SECURE_COOKIE = process.env.REACT_APP_USE_SECURE_COOKIE !== "false";
 export const DEBUG = process.env.REACT_APP_DEBUG === "true" || false;
```
hmpf, line 5: In prod, this entire file is replaced. We still don't have a good solution for this. A matter for another PR.

hmpf (changes requested):
There are some documentation nits.

README.md
lunkwill42: Yes. I've been editing both reStructured Text docs and Markdown docs today, so apparently I got confused about the markup :-D

docker/api/Dockerfile
lunkwill42: I don't, but the `dev` settings which `dockerdev` is based on does. Removing dotenv from argus backend should probably be posted as an issue there.

src/config.tsx
lunkwill42: Indeed, and I'm not sure we *can* have a better solution, since we only deploy static files in production (though I'm sure some crazy people run Node server directly in production and can use the environment for config).

I've mentioned this bit in #180

docker/api/docker-entrypoint.sh
lunkwill42: Should I point to the argus backend docs, or copy&paste the argus backend docs? :-)

hmpf (approved):
I've added an issue https://github.com/Uninett/Argus-frontend/issues/181 for the README, it needs improving in general.

PR #182, opened 2020-12-17
Implement automatic updating/refetching of incidents

jorgenbele:


lunkwill42 (approved):
:+1:

Once we can make websockets work consistently, and with filters, this needs to change into a fallback strategy for when the frontend detects that the websocket strategy isn't working

PR #184, opened 2020-12-17
Add route for showing incident by canonical url

### LINKED ISSUE #148: Provide canonical URLs to access the incident details "dialog"
It should be possible for 3rd party tools, like ticketing systems, to link back to the Argus details page for a single Incident, which means that each incident needs a canonical URL to enable access to this.

Since the Argus UI is a single-page-application, the details dialog/modal is currently only available by manually clicking through the interface.

jorgenbele:
This PR implements the required functionality of #148.

It includes the following:
- [X] Split the incident view => incident view, incident detail component, incident toolbar component
- [X] Make URLs on the form: /incident/<incident pk> open a view consisting of only the incident details

This required refactoring of <Header/> component usage. 

Closes #148

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,6 +16,21 @@ import auth from "./auth";
 import { ThemeProvider } from "@material-ui/core/styles";
 import { MUI_THEME } from "./colorscheme";
 
+import Header from "./components/header/Header";
+
+// eslint-disable-next-line
+const withHeader = (Component: any) => {
+  // eslint-disable-next-line
+  return (props: any) => (
+    <>
+      <header>
+        <Header />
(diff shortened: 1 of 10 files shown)


lunkwill42 (approved):
Still not experienced enough in React to make any useful code review here, but I have tested and found the functionality works as described.

The only thing I would add, I think, is for the incident description to be displayed more prominently when the incident is opened on a separate page.

* E.g. a `<h1/>` header or similar with the description as the page "title"
* Putting the description in the actual HTML document title, so that the page can be correctly identified among multiple tabs (and if bookmarks are saved, they will be titled appropriately)

*

PR #185, opened 2020-12-17
Create new header/top navigation bar

jorgenbele:
This PR introduces a completely redesigned header/navbar implemented by using AppBar from MUI. This resolves some of the problems with the hand-written CSS one that was previously in use: 1. items didn't align properly, 2. didn't behave well when zooming.

Based on #184, so merging this will merge that PR as well.

2020-12-17-205304_1244x241_scrot


lunkwill42 (approved):
:+1:

PR #187, opened 2020-12-20
Minor improvements to incidents toolbar, new feature of an drop-down with additional setting 

jorgenbele:
This PR includes some minor improvements to the incidents filter toolbar (rewriting to remove use of Grid), as well as introducing a drop-down that is supposed to be used for useful filter settings that are not needed to be visible at all times.

Currently the only setting in this drop-down is the auto update setting, but there will probably be use for it.

2020-12-21-002909_1236x176_scrot

2020-12-21-002922_1233x271_scrot

The previous PRs should be merged before this one.


lunkwill42 (approved):
Looks good to me :-) :+1:

Not merging, since this is draft PR...

PR #192, opened 2020-12-22
Redesign login view and split it into components

jorgenbele:
This isn't the best designed login view, but this is a lot better as a base to build a new and nicer one than the previous.
2020-12-22-120204_1285x1122_scrot

### DIFF
--- a/src/components/buttons/OutlinedButton.tsx
+++ b/src/components/buttons/OutlinedButton.tsx
@@ -1,0 +1,54 @@
+import React from "react";
+
+import { createStyles, makeStyles, Theme } from "@material-ui/core/styles";
+import Button, { ButtonProps } from "@material-ui/core/Button";
+
+import classNames from "classnames";
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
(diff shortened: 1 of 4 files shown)


lunkwill42 (changes requested):
Looks fine by me, however, it seems like the "user icon" in the top left corner isn't updated when I log in. It says "unknown", until I click, then it updates to show the user I am currently logged in as...

also, I find the contrast between black text in the form fields and the dark blue background to be too small. I think the text should probably be white as well

jorgenbele:
> Looks fine by me, however, it seems like the "user icon" in the top left corner isn't updated when I log in. It says "unknown", until I click, then it updates to show the user I am currently logged in as...

I'm also aware of this, but that isn't related to this PR. It's a bug with how the the user data is recieved. It should still be fixed of course.

> 
> also, I find the contrast between black text in the form fields and the dark blue background to be too small. I think the text should probably be white as well

Yes, I'm aware, and I will fix it, I just didn't find the correct way to set the color in time.

lunkwill42 (approved):

lunkwill42:
Fine, I will accept these changes and report the contrast issue as a bug :-)

PR #193, opened 2020-12-22
Feat better realtime integration & UI changes

jorgenbele:
This PR tries to improve the realtime integration in Argus:

1. Seperate out the Websockets logic from incidents table to a seperate service class. 
2. Make incidents that are updated (by user modification or by realtime update) stay visible for 5 seconds before they disappear IF they are supposed to disappear. --- some exceptions that needs to be fixed.
3. Add border to the left-side of the with a color representing the state

Relies on previous PRs, at least #191.

### DIFF
--- a/src/components/incident/styles.tsx
+++ b/src/components/incident/styles.tsx
@@ -43,8 +43,7 @@ export const useStyles = makeStyles((theme: Theme) =>
       color: WHITE,
     },
     tableRow: {
-      // background: theme.palette.success.light,
-      borderLeftWidth: "20px",
+      borderLeftWidth: "10px",
       borderLeftStyle: "solid",
       borderLeftColor: "transparent",
     },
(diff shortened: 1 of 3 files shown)


lunkwill42 (commented):
Question about 7b03a4bad7ad2c4fc799d7fdd3f59355e10d4db9:

To me, this appears to work if I use the frontend to close or ack an incident - but not if the incident is closed by the backend (i.e. the color band changes, but the incident line remains present).

Is it just me, or can you reproduce it?

jorgenbele:
> Question about [7b03a4b](https://github.com/Uninett/Argus-frontend/commit/7b03a4bad7ad2c4fc799d7fdd3f59355e10d4db9):
> 
> To me, this appears to work if I use the frontend to close or ack an incident - but not if the incident is closed by the backend (i.e. the color band changes, but the incident line remains present).
> 
> Is it just me, or can you reproduce it?

I will try to reproduce it now

jorgenbele:
I can reproduce it, but I don't think this is new.

lunkwill42:
> I can reproduce it, but I don't think this is new. I think I should spend some time trying to implement proper filtering (which is the problem here) in the frontend, because that haven't been attempted since summer.

We should probably have proper design discussion around this in January then, I think. But this is okay for now :-)

lunkwill42 (approved):

PR #207, opened 2021-01-12
Add functionality for using predefined filters in incident filter toolbar

jorgenbele:
Closes #173

PR #210, opened 2021-01-15
Bugfix realtime this connect

### LINKED ISSUE #198: Bug: this.connect() is not a function error occurs when websocket fails
**Describe the bug**
When the WebSocket connection fails there is thrown an error due to a call to `this.connect()` in the realtime service. For some reason this doesn't work - probably due to some black javascript magic. 

**To Reproduce**
1. Force the active realtime websocket connection to fail
2. See the debug log/popup

**Expected behavior**
The call to this.connect() should work, and and there should be an attempt to get a new websocket connection.

jorgenbele:
Closes #198


hmpf (changes requested):
This has the same problem as #207. Please add the missing classnames, I cannot run the frontend locally otherwise.

hmpf (approved):
Works after manual testing.

PR #218, opened 2021-01-20
Keep track of user in global app context

### LINKED ISSUE #199: Bug: avatar/profile in header is missing data before first interaction
**Describe the bug**
When you login the avatar/profile component in the header will say unknown before the first interaction with the header.

**To Reproduce**
1. Go to login view
2. Login using FEIDE
3. Look in the top right corner

**Expected behavior**
The expected behaviour is that the profile name should be displayed, not Unknown.

**Additional context**
The reason this happens is that the header component gets the name from localStorage (legacy reasons), something that only happens whenever react decides to do so. If the user isn't set in localStorage before react decides to rerender, it will get the default value of unknown.

jorgenbele:
This PR adds information about the user to the global app context, and replaces some (not all) uses of user information with this. 

Closes #199

### DIFF
--- a/src/components/header/Header.tsx
+++ b/src/components/header/Header.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useContext } from "react";
 import "./Header.css";
 import auth from "../../auth";
 import { Link, useHistory, withRouter } from "react-router-dom";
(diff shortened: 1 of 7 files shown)


hmpf (commented):
I'll test locally as well.

hmpf (approved):

PR #220, opened 2021-01-21
Show datetimes as humanized localtime ISO

### LINKED ISSUE #137: Use consistent DateTimes
The datepicker for setting an ack's expiry date is hardcoded to US dates.
It is hard-coded to MM/dd/yyyy. It should use whatever the locale uses. (Preferably by something looking up the locale's date-format and storing it in a variable so that it can be overridden by the members of the Church of ISO 8601 later =) )

hmpf:
Closes #137

### DIFF
--- a/src/components/incidenttable/FilteredIncidentTable.tsx
+++ b/src/components/incidenttable/FilteredIncidentTable.tsx
@@ -13,6 +13,8 @@ import { useApiPaginatedIncidents } from "../../api/hooks";
 import { DEFAULT_AUTO_REFRESH_INTERVAL } from "../../config";
 import { LOADING_TEXT, ERROR_TEXT, NO_DATA_TEXT } from "../../constants";
 
+import { formatTimestamp } from "../../utils";
+
 type PaginationCursor = {
   next: string | null;
   previous: string | null;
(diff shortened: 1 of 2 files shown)


src/components/incidenttable/FilteredIncidentTable.tsx
```diff
@@ -235,7 +237,7 @@ const FilteredIncidentTable: React.FC<FilteredIncidentsTablePropsType> = ({
         paginationComponent={paginationComponent}
       />
       <p>
-        Last refreshed {lastRefresh === undefined ? "never" : lastRefresh.toLocaleTimeString()}, {autoUpdateText}
+        Last refreshed {lastRefresh === undefined ? "never" : formatTimestamp(lastRefresh).slice(11)}, {autoUpdateText}
```
jorgenbele, line 240: Why do you use slice(11) here?

jorgenbele:
Before you submit a PR you should run `yarn format` or equivalent using npm to reformat the code.

src/components/incidenttable/FilteredIncidentTable.tsx
hmpf: I'm not using the entire format. Microseconds are unwanted,as is timezone.

hmpf:
Closed by #240

Issue #205, opened 2021-01-06
Create a filter from the front page instead of the filter list


hmpf:
Needs: A place to input the filter name.

hmpf:
Closed by #223

PR #223, opened 2021-01-25
Feat filter from frontpage

jorgenbele:


hmpf:
The following screenshot shows when no filter is chosen:

image

I very much like this premise. Deletion is hidden behind the gear which is smart.

The last screenshot shows when a filter is chosen:

image

I do think that the difference between "No filter chosen" and "Filter chosen" is too subtle. Decreasing contrast when a filter is not chosen might be enough. Also, it might be that having the filters in alphabetical order is better (now it seems to be oldest first), though we should get more opinions on that.

Considering the line-breaks when applying my "Everything"-filter, we might have to figure something out in that case (but that is a _different_ issue, not to be fixed in this one!)

Finally, since we do not support storing openness/ackedness in a backend-filter, it is possible to create a filter that does nothing (you click "Save" without having selected a source or tags), I think the correct solution for this is to add openness/ackedness to the filter asap. Though, this code should probably not be merged until the backend-filter has been improved.

jorgenbele:
> I do think that the difference between "No filter chosen" and "Filter chosen" is too subtle. Decreasing contrast when a filter is not chosen might be enough. Also, it might be that having the filters in alphabetical order is better (now it seems to be oldest first), though we should get more opinions on that.

I can understand that the difference might be subtle, but I don't really see any downsides to this. I think adding a cyan "selected" background to the filter drop-down when something is selected fits with design language used currently, so I will add that in.

> Considering the line-breaks when applying my "Everything"-filter, we might have to figure something out in that case (but that is a different issue, not to be fixed in this one!)

Yeah, this could be a bit tricky. One option is to have a "expand" button in the source- and tag selection components if there are more than a set amount.

> Finally, since we do not support storing openness/ackedness in a backend-filter, it is possible to create a filter that does nothing (you click "Save" without having selected a source or tags), I think the correct solution for this is to add openness/ackedness to the filter asap. Though, this code should probably not be merged until the backend-filter has been improved.

Here I don't really see a problem. The frontend doesn't currently handle "openness/ackedness" as if it was part of a "backend filter", but that wouldn't be hard to implement when backend support is added. Are empty filters a problem?

hmpf:
> > I do think that the difference between "No filter chosen" and "Filter chosen" is too subtle. Decreasing contrast when a filter is not chosen might be enough.
> 
> I can understand that the difference might be subtle, but I don't really see any downsides to this. I think adding a cyan "selected" background to the filter drop-down when something is selected fits with design language used currently, so I will add that in.

Thinking some more, the problem might be the word "None". To me, that looks like a name of an actual filter. "Not saved" or just an empty box might be better.

jorgenbele:
I will remove the "None" text, and just let it be empty then. I'll still explore the ideas above though

jorgenbele (commented):
> Thinking some more, the problem might be the word "None". To me, that looks like a name of an actual filter. "Not saved" or just an empty box might be better.

This commit removes the "None" sentinel and instead will display nothing when no filter is selected. In order to clear the filter selection an additional button was added. I think it fits in quite nicely, because the button is placed where the "create filter" button is placed when a filter is not selected.

jorgenbele:
This closes #205

jorgenbele:
Relies on the PRs of #227, #226 and #225. For now it includes the commits of them so that it's easily testable. Needs to be rebased before merging.

hmpf:
Thx for the last comment. I'll merge in the many small approved dependencies locally and play a little, then nudge you when it is time to rebase this one and the other monster.

lunkwill42 (commented):
I really like the direction this is taking.

I did a quick test, and I found one **confusing bug**:
* When I save the active filter expression to a new (or existing) filter, all filters are then removed from the incident list, and I see everything again...

jorgenbele:
Yeah, it makes sense that you observe that. I haven't considered it a bug, but I could make it change to the saved filter on save. Is that better do you think?

hmpf:
Vidar S. was also puzzled by what happens when saving/updating a filter so yes. needs changing. Otherwise, we are go!

hmpf:
In order to deploy a review I pulled the branch down to this repo. Probably best if you fix your branch in your repo though if  it is to be the final variant.

jorgenbele:
I've fixed the issue of not setting the selected filter on save as/creation. I also made it close the modal/dialog when creating/saving is successful. Should be good to go now.

Issue #209, opened 2021-01-13
Revisit incident pagination


katsel:
Currently, the number of incidents displayed is set to a fixed number set by the user (via dropdown-menu).

This has certain unpleasant side-effects with real-time viewing:

- New incidents coming in will exceed the page size limit
  - Page will overfill, or the last incidents get dropped
- Incidents may change their status real-time and not match the filter any longer, so the incident number falls behind the page size limit
  - Opposite as above
 
All these effects are to be expected when dealing with pagination and real-time data. This rises the question whether pagination is a useful approach here at all, and what alternatives are better suited to real-time data.

Possible alternatives include endless scrolling and a "Load more" button.
These are also more user friendly, as they take unnecessary decisions away from the user, providing a sleeker and more predictable experience when using Argus.

katsel:
Setting this issue to "blocked" as implementation details may depend on the outcome of Uninett/Argus/issues/250.

Discussion is welcome nonetheless.

hmpf:
We use a different type of pagination in the new frontend, and users have not asked or noticed this problem, closing.

PR #229, opened 2021-01-28
Feat realtime list

### LINKED ISSUE #200: Bug: incidents retrieved using websockets doesn't filter properly
**Describe the bug**
When the user has selected a filter, and an incident that does not match that filter is either added or modified, it will still display that incident.

**To Reproduce**
1. Go to /incidents
2. Select a filter
3. Generate an incident or modify an existing one such that it does not match the current filter
4. See error, the incident appears even when it shouldn't

**Expected behavior**
The filtering capabilities should still work as it does when not using realtime.

### LINKED ISSUE #201: Bug: disabling realtime doesn't work correctly
**Describe the bug**
Even though there are settings to switch between realtime, interval and never, you can't really ever disable realtime.

**To Reproduce**
1. Try to disable realtime (config or in frontend)
2. Observe that nothing changes

**Expected behavior**
The realtime websockets connection should only start when it is the enabled one.

### LINKED ISSUE #217: Sources filter on front page is not applied to new incidents when using Auto Update.
**Describe the bug**
In Argus frontpage, when Argus is updated with new incidents and you have a Sources filter applied, then it seems that new incidents are not filtered. For instance with the screen shot example, where incidents from uninav and lokalnav are shown even though the Sources only includes sias-vk, khio-vk and unit-vk. This is because these Incidents have occurred after I applied these filters. 

**To Reproduce**
1. Go to Argus frontend
2. Apply a filter in 'Sources' for instance only 'khio-vk.c.uninett.no'
3. Wait for an incident appearing from another source, for instance 'sias-vk, uninav, lokalnav, unit-vk'. 
4. Observe that this incident occurs in the list of incidents. 

**Expected behavior**
I expect that new incidents also gets filtered based on the Sources or Tags selected. 

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 87.0.4280.141 (Official Build) (64-bit)

jorgenbele:
This PR works on the realtime features. It switches to using contexts for sharing state, and adds some fixes to the RealtimeService, as well as properly handling switching between realtime/interval. +++

Closes #201
Closes #200 
Closes #217

Interacts with #209 by making the decision to not use paging (it's just removed for now). This can probably be undone, because the improved internal architecture makes such changes easier. 

This PR requires rebasing, but because the # of PRs keep increasing I will not spend time doing that everytime I change something. This is based on all previous PRs, even those that are not merged.

NOTE: Not infinitely loading yet, but will be added

### DIFF
--- a/src/api/actions.tsx
+++ b/src/api/actions.tsx
@@ -1,5 +1,5 @@
 import React, { useContext } from "react";
-import api, { Filter, FilterSuccessResponse } from "../api";
+import api, { Filter, FilterSuccessResponse, Incident, IncidentsFilterOptions } from "../api";
 import { InitialStateType, ActionsType } from "../contexts";
 import {
   createFilter as createFilterAction,
@@ -7,6 +7,10 @@ import {
   modifyFilter as modifyFilterAction,
   loadAllFilters as loadAllFiltersAction,
 } from "../reducers/filter";
+
+import { IncidentsFilter } from "../components/incidenttable/FilteredIncidentTable";
+import { Tag } from "../components/tagselector";
+import { IncidentsType, IncidentsStateType, useIncidentsContext } from "../components/incidentsprovider";
 
 import { AppContext } from "../contexts";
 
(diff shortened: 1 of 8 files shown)


hmpf:
I'll look at this one once more of the others have been merged.

hmpf (approved):
Tested it. With realtime off, it takes longer to redraw the list when changing filters.

This points to: Maybe we should have a dedicated review site with live updates of data in order to only test new frontend things?

jorgenbele:
> Tested it. With realtime off, it takes longer to redraw the list when changing filters.

That is interesting, because there are no changes to the non-realtime component in this PR. Which means that the following https://github.com/Uninett/Argus-frontend/blob/0713117386612785ce6e8d66b2d92cdf50752785/src/views/incident/IncidentView.tsx#L17, is where this extra time is spent. I don't understand how this takes any significant amount of time.

> This points to: Maybe we should have a dedicated review site with live updates of data in order to only test new frontend things?

I'm not sure if this is what points to it, but yes I've requested that in the past. A "develop" branch that i can merge/push to at will would be useful for this.

PR #232, opened 2021-01-30
Feat table toolbar

### LINKED ISSUE #222: The possibilities to update multiple events at the same time.
It would have been really nice to update multiple events with the same data at the same time. 

For instance if I need to set similar Acknowledgement on or close multiple events with the same message, I today need to go in to the events one at a time and update it. Then it would have been good to be able to doing this in one action.

jorgenbele:
This PR adds the functionality requested in #222 

2021-01-30-235704_2556x762_scrot

Closes #222 

Relies on the previous still open PRs, so they should be merged before this one.


jorgenbele:
This was merged (by mistake might I add) in another PR

PR #233, opened 2021-01-31
Realtime improvements & upgrade of table

jorgenbele:
This PR (draft) does the following:
 1. It improves on the RealtimeService, such that the components has better access to state.
 2. It adds a better way of dealing with frontend filtering, and shares it between the realtime incident component and the interval incident component.
 3. Fixes some useEffect's, that previously didn't handle dependencies correctly

As is tradition, this needs to be rebased when the other PRs are merged.

### DIFF
--- a/src/api/actions.tsx
+++ b/src/api/actions.tsx
@@ -1,5 +1,13 @@
 import React, { useContext } from "react";
-import api, { Filter, FilterSuccessResponse, Incident, IncidentsFilterOptions } from "../api";
+import api, {
+  Filter,
+  FilterSuccessResponse,
+  Incident,
+  IncidentsFilterOptions,
+  Event,
+  AcknowledgementBody,
+  Acknowledgement,
+} from "../api";
 import { InitialStateType, ActionsType } from "../contexts";
 import {
   createFilter as createFilterAction,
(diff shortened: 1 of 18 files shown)


hmpf (changes requested):
Running this when websockets doesn't work properly (redis isn't reacahable for instance) leads to an endless loop of a message box "Realtime connecting" > "Realtime opened" > "Realtime closed" popping and repopping. Distracting.

jorgenbele:
> Running this when websockets doesn't work properly (redis isn't reacahable for instance) leads to an endless loop of a message box "Realtime connecting" > "Realtime opened" > "Realtime closed" popping and repopping. Distracting.

@hmpf This should only be the case when you have selected realtime right? Because I think it makes sense to have an annoying message when something doesn't work and you have to take manual measures to either fix the problem, or select interval/never.

hmpf:
When realtime update is not set (both "never" and "interval" are grey, which seems to be the start state), this happens. When selecting "Never", this still happens. When selecting "interval". this still happens. That can't be right, surely? Also, how is a user to know how to fix this (if "never" or "Interval" fixes things)? Also, the "disconnect"-message flashes by so fast that it is easy to miss.

jorgenbele:
> When realtime update is not set (both "never" and "interval" are grey, which seems to be the start state), this happens. When selecting "Never", this still happens. When selecting "interval". this still happens. That can't be right, surely? Also, how is a user to know how to fix this (if "never" or "Interval" fixes things)? Also, the "disconnect"-message flashes by so fast that it is easy to miss.

I can confirm that this not the intended functionality. I think I know what causes this, and suspect that this wouldn't happen if you disabled realtime in the configuration. I will investigate.

jorgenbele:
@hmpf I've done some changes. Can you check if you still observe these problems?

hmpf:
Willdo.

hmpf:
Seems solved, merging.

hmpf (approved):
This does seem to contain much (all?) of the functionality of #232 but since we do want that functionality I'm merging this one as is in the interest of saving some time.

Issue #237, opened 2021-02-19
Bug: When the backend cannot be reached (network problem) nothing sensible is shown


hmpf:
**Describe the bug**
If the frontend cannot access the backend after login, either everything hangs with no form of alert, or an "Unhandled Rejection" pops up.

**To Reproduce**
1. Log in to a frontend that uses a backend (via BACKEND_URL) that can be easily shut off, like for instance localhost 
2. Turn off the backend
3. See problem (attached screenshot)

**Expected behavior**
Show some error-message that explains that the backend is unreachable.

**Screenshots**

Automated error: counter (1 of 5) slowly increases if "interval" is on.
![image](https://user-images.githubusercontent.com/11956096/108476925-e2280f80-7292-11eb-9ea6-96f3f4b7c4bd.png)

If we click that away:
![image](https://user-images.githubusercontent.com/11956096/108477182-30d5a980-7293-11eb-8d0c-08acf2bf49af.png)

**Additional context**
If the backend is unreachable prior to/during login, the browser itself will complain, so no need to cover that angle.

jorgenbele:
Closing, because it was solved in #239

PR #239, opened 2021-02-19
Feat network problem indicator

jorgenbele:


jorgenbele:
Closes #237

jorgenbele:
Here is what it looks like when the network connection to the backend is detected: 
image

lunkwill42 (commented):
Tested superficially, and it looks :+1: 😄 

However, in dev mode the error keeps popping up as a huge exception I need to dismiss before I can see the friendly error message from Argus-frontend. 

Assuming this behavior is intentional, and will not bother regular users?

I'll leave the final decision up to @hmpf :)

jorgenbele:
> Tested superficially, and it looks +1 smile
> 
> However, in dev mode the error keeps popping up as a huge exception I need to dismiss before I can see the friendly error message from Argus-frontend.
> 
> Assuming this behavior is intentional, and will not bother regular users?
> 
> I'll leave the final decision up to @hmpf :)

Yes, the exception popup in debug mode is expected, and happens whenever any unhandled exception is thrown AFAIK. I don't think there is any way to remove this without negative side-effects. You could probably modify your local development environment, but in testing these error messages are useful.

hmpf (approved):

PR #240, opened 2021-02-19
Allow customizable timestamps

jorgenbele:
This implements functionality of #220 as well as:
1. Removing use of moment.js (deprecated/no longer maintained), replacing it with date-fns
2. Adding ability to specify format in configuration file, and also conditionally enable seconds/timezone-offset by passing an options dict to `formatTimestamp()`

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,9 @@
 import React from "react";
 
-import "./variables.css";
-import "./colorscheme.css";
-
-import { Route, Switch, useHistory } from "react-router-dom";
+import { Route, Switch } from "react-router-dom";
 import { ProtectedRoute } from "./protected.route";
 
+// Views
 import IncidentDetailsView from "./views/incident/IncidentDetailView";
 import IncidentView from "./views/incident/IncidentView";
 import LoginView from "./views/login/LoginView";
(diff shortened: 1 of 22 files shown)


jorgenbele:
Closes #137

jorgenbele:
Closes #220

lunkwill42 (changes requested):
This looks lovely on the incident list! :+1:

My only feedback is that I would actually prefer including seconds in the timestamp when viewing the event details dialog (after all, details should be detailed!) :-)

jorgenbele:
> This looks lovely on the incident list! +1
> 
> My only feedback is that I would actually prefer including seconds in the timestamp when viewing the event details dialog (after all, details should be detailed!) :-)

I've set the `withSeconds` flag for all uses in the IncidentDetails. Does that work?

hmpf (approved):

hmpf:
> > This looks lovely on the incident list! +1
> > My only feedback is that I would actually prefer including seconds in the timestamp when viewing the event details dialog (after all, details should be detailed!) :-)
> 
> I've set the `withSeconds` flag for all uses in the IncidentDetails. Does that work?

Yep, and I can verify: it looks good, so I'm merging.

Issue #208, opened 2021-01-13
Persist incident page size between app reloads


lunkwill42:
The default page size for incidents is currently 10 (which, IMHO, is optimistically low). I can select among various other page sizes from a dropdown. However, my selection is not persisted when I reload the entire app, which is quite annoying.

**Feature request**
Persist the user's choice of page size into their browser's local storage. This way, I will always have a predictable page size.

katsel:
Please also

- [ ] increase default incident size to 20 or 25

lunkwill42:
Actually, this might become obsolete, as @jorgenbele is working to replace paging with infinite scroll, I think..

jorgenbele:
> Actually, this might become obsolete, as @jorgenbele is working to replace paging with infinite scroll, I think..
> 
> If you have a PR or issue reference, please add it here..

I have only planned on using infinite scroll when realtime is used, keeping paging when interval/never is used.

PR #241, opened 2021-02-19
Feat keep preferences in localstorage

jorgenbele:
This PR introduces features that allow for persistence of values in localStorage. It does this by copying parts of contexts/states on every change to them, and then attempting to load from localstorage on initialization. 

Currently implements storage of the following:
1. Page size (for Interval)
2. Selected filter (including open/closed, acked/unacked, and If realtime/interval/never is selected).

It also stores a timestamp along side with the data, which might come in useful during eventual debugging or future improvements.

### DIFF
--- a/src/components/filterprovider.tsx
+++ b/src/components/filterprovider.tsx
@@ -1,4 +1,4 @@
-import React, { useReducer, useContext, createContext } from "react";
+import React, { useEffect, useReducer, useContext, createContext } from "react";
 
 import { IncidentsFilter, AutoUpdate } from "../components/incidenttable/FilteredIncidentTable";
 import { Filter } from "../api";
@@ -6,6 +6,10 @@ import { Tag, originalToTag } from "../components/tagselector";
 
 // Store
 import { ActionMap } from "../reducers/common";
+import { SELECTED_FILTER } from "../localstorageconsts";
+
+// Utils
+import { fromLocalStorageOrDefault, saveToLocalStorage } from "../utils";
 
 export type SelectedFilterStateType = {
   // Optionally selected using drop-down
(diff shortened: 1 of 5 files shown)


jorgenbele:
Closes #208

lunkwill42 (commented):
ATM I'm only getting this error when testing this branch in my browser:

```
Failed to compile

./src/components/incident/IncidentFilterToolbar.tsx
Module not found: Can't resolve '../../localstorageconsts' in '/app/src/components/incident'
```

I'm going to assume this is more npm bullshittery that I don't understand simply because I don't know the NPM ecosystem, so I will try some more...

hmpf:
I get the same. There is no file "localstorageconsts.tsx" in `./src`. `IncidentFilterToolbar.tsx` only imports `DROPDOWN_TOOLBAR` from this file, and the string `DROPDOWN_TOOLBAR` does not exist anywhere else in the tree. I guess the file was overlooked?

Maybe `DROPDOWN_TOOLBAR` could be housed in `config.tsx` instead of a new file, but in any case, either:

1. an update of the README is needed, if we need to add a file that should not be in version control
2. if it may be in version control it needs to be added to the PR

src/components/incident/IncidentFilterToolbar.tsx
```diff
@@ -43,6 +43,10 @@ import api, { Filter, IncidentMetadata, SourceSystem } from "../../api";
 // Config
 import { ENABLE_WEBSOCKETS_SUPPORT } from "../../config";
 
+// Utils
+import { saveToLocalStorage, fromLocalStorageOrDefault } from "../../utils";
+import { DROPDOWN_TOOLBAR } from "../../localstorageconsts";
```
hmpf, line 48: This file does not exist.

src/components/filterprovider.tsx
```diff
@@ -1,11 +1,15 @@
-import React, { useReducer, useContext, createContext } from "react";
+import React, { useEffect, useReducer, useContext, createContext } from "react";
 
 import { IncidentsFilter, AutoUpdate } from "../components/incidenttable/FilteredIncidentTable";
 import { Filter } from "../api";
 import { Tag, originalToTag } from "../components/tagselector";
 
 // Store
 import { ActionMap } from "../reducers/common";
+import { SELECTED_FILTER } from "../localstorageconsts";
```
hmpf, line 9: There's no no complaint here so I guess `IncidentFilterToolbar.tsx` is processed before `filterprovider,tsx`.

hmpf (dismissed):
`localstorageconsts.tsx` is missing from the repo.

jorgenbele:
I've added the missing file. It doesn't ever need to be changed, so it doesn't make sense to pollute the config file with it.

lunkwill42 (approved):
:+1:

PR #249, opened 2021-03-25
Remove docker-compose restart directives

lunkwill42:
Because:

- The docker-compose config is for development environments.
- There is no need for automatic restarting of containers in a dev env.
- This causes manuallt stopped containers to restart whenever the Docker  daemon is restarted (e.g on every reboot)


jorgenbele (approved):
Seems very reasonable to me. :+1:

PR #254, opened 2021-04-28
Add docker-compose setup for testing production-build of frontend

jorgenbele:


lunkwill42:
This is a good idea, but I don't particularly like the pollution of the top level directory this results in. I would have liked most of the added files to go in the `docker` directory, or a sub-directory thereof

PR #266, opened 2021-06-17
Test/login

hermar98:


lunkwill42 (approved):
Not being great with either TypeScript, React nor Jest, I can still sort of grok this :)

It seems to be a good start on the testing framework and GitHub Actions - hoping for a walkthrough IRL later today!

👍

PR #251, opened 2021-04-21
Upgrade filter

jorgenbele:
Sooo, this PR contains the following:
1. Refactoring to make the development process better
2. Use CSS for truncating text instead of javascript
3. useUser() hooks
4. Supports new API

PR #268, opened 2021-06-21
Upgrade filter fix bugs

hermar98:
Replaces #251. From @jorgenbele's description:

1. Refactoring to make the development process better
2. Use CSS for truncating text instead of javascript
3. useUser() hooks
4. Supports new API


lunkwill42 (commented):
I assume this is here to replace #251 

I will run a perfunctory manual test, but the commit log seems to contain fixup commits. I.e., some rebasing is probably need to clean up the logs.

If you're not comfortable with interactive rebasing in Git, hit me up tomorrow, and we can look at it together.

lunkwill42 (changes requested):
> I will run a perfunctory manual test,

Looks fine and dandy to me :)

Now it's just down to slightly cleaning up the commit logs via rebase..

lunkwill42 (approved):
Great work, @jorgenbele  and @hermar98 !

PR #269, opened 2021-06-21
Tests/incidents toolbar

podliashanyk:


lunkwill42 (commented):
This is pretty huge, but the bulk of it seems to be lots of generated coverage reports. I think we should do a short IRL session to whittle it down to something more manageable.

Please remind me at tomorrow's standup :)

lunkwill42 (changes requested):
After today's session, we concluded:

* The `coverage/` directory seems to consist only of generated reports. These don't belong in version control, so the directory contents should be excluded in `.gitignore`, and be stripped from this PR (git rebase is your friend).

In addition, we decided it's a good idea to put `package-lock.json` in `.gitignore` on master as well, so that we only force push changes to this file at specific intervals (like, a new release version). If either @hermar98 or you fix this on master, this PR should probably be rebased onto the new master and the `package-lock.json` changes stripped as well.

lunkwill42 (approved):
Nice work :)

Just one question: I see the testing step of the workflow logs contain a lot of these messages:
https://github.com/Uninett/Argus-frontend/pull/269/checks#step:6:10

> (node:1836) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 3)

Not sure what it means, but it looks ominous...

lunkwill42 (approved):

PR #270, opened 2021-06-23
Fix invalid time bug in Safari

podliashanyk:
Replaces Date() constructor with date-fns' parseISO(). This solution is chosen over converting an invalid timestamp string into a valid one, since the latter involves a more complex refactoring and is less 'universal' in implementation.


lunkwill42 (approved):
Nicely done! :)

PR #273, opened 2021-06-28
Add ticket url bulk

### LINKED ISSUE #265: Set ticket url as a bulk operation
Today it is only possible to add a ticket ID to an incident by clicking the incident and adding it. 

I would like to be able to put the same ticket ID on multiple incidents the same way as it is possible to ACK multiple incidents at the same time. 

image

podliashanyk:
Adds "Add Ticket" button for bulk adding ticket url to selected incidents. Tests implemented functionality. Closes #265

### DIFF
--- a/src/components/incident/AddTicketUrlSignOffAction.tsx
+++ b/src/components/incident/AddTicketUrlSignOffAction.tsx
@@ -1,0 +1,34 @@
+import React from "react";
+
+import SignOffAction, { SignOffActionPropsType } from "./SignOffAction";
+
+type AddTicketUrlPropsType = {
+  onAddTicketUrl: (url: string) => void;
+  signOffActionProps?: Partial<SignOffActionPropsType>;
+};
+
+const AddTicketUrl: React.FC<AddTicketUrlPropsType> = ({
+                                                       onAddTicketUrl,
+                                                       signOffActionProps = {},
+                                                     }: AddTicketUrlPropsType) => {
+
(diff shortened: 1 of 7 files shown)


src/components/incident/AddTicketUrlSignOffAction.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <SignOffAction
+      dialogTitle="Add ticket URL"
+      dialogContentText="Write an URL of an existing ticket"
+      dialogSubmitText="Submit"
+      dialogCancelText="Cancel"
+      dialogButtonText="Add ticket"
+      title="Add ticket url"
+      question="Are you sure you want to add ticket url?"
```
hmpf, line 27: Would this be better as "Are you sure you want to add a ticket url?" I'm not sure, I'm so used to getting rid of the/a in commit messages :)

hmpf (commented):
Feels a bit slow. I'm testing without realtime, and the page does not redraw until the expected time (every 30 seconds), which makes it seem like something didn't work. Might be several ways of solving this.

There's also the matter of whether one should hide the multi-select bar on success, or deselect the selected tickets, since with ack and close, depending on what filter is set, the changed tickets disappear from the list even while the multi-select bar remains, with "N tickets" still marked.I think we need to look at this together and decide what is the assumedly best/least confusing UI.

hmpf (approved):

PR #274, opened 2021-06-28
Incident severity levels

hermar98:
The frontend now shows the severity level for each incident, both in the incident list and in the details dialog. The names and colors of the severity levels is only a suggestion, and can easily be changed later.

Here are some screenshots of the changes that has been made:

## Incident list:
image

## Incident details dialog:
image

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -388,6 +388,9 @@ class ApiClient {
       if (filter.filter.stateful !== undefined) {
         params.push(`stateful=${filter.filter.stateful}`);
       }
+      if (filter.filter.maxLevel !== undefined) {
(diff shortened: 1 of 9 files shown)


hermar98:
Now I have also implemented filtering on severity level in the frontend. This is done by selecting max level in the dropdown in the incidents toolbar. If for example max level is selected to be "3 - moderate", incidents with severity level 1, 2 and 3 will be shown. The max level will also be stored if in a filter when it is created. The feature looks like this:

Skjermbilde 2021-07-26 133820
Skjermbilde 2021-07-26 133848

This pull request now also fixes bug #286.

hmpf (approved):

hmpf:
Needs UI-polish, but that is another issue/PR.

PR #280, opened 2021-06-30
Shows success/error snackbars on all bulk operations on incidents

### LINKED ISSUE #278: Provide feedback on operation result when modifying an incident
The only feedback user gets on operation flow when modifying an incident (ack/close/reopen/add ticket) is that appropriate chips change color/a ticket label appears/incident disappears from the list due to the filter restrictions. A more complete feedback would be to display a snackbar that disappears after X seconds. Solution with snackbars enables providing feedback on erroneous operations/errors from server. 
Snackbar examples: 
Screenshot 2021-06-29 at 10 35 06
Is linked to #124

podliashanyk:
Closes #278


hmpf (approved):
Looks good, tests good.

PR #281, opened 2021-06-30
Shows error cause to user

### LINKED ISSUE #124: Polish: The error-message from 400 Bad Request aren't shown to the enduser
The backend frequently sends a json-body with its 400 Bad Request-replies. It is better UX to show these than only "400 Bad Request" when something breaks. (It also speeds up debugging!)

podliashanyk:
Error cause is now shown to user in JSON format. Not the prettiest solution, but definitely the simplest one to implement. Showing cause string directly involves going in a rabbit hole of checking each endpoint, and writing "adapter"-code on frontend.
Closes #124

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -44,6 +44,7 @@ import auth from "../auth";
 import { ErrorType, debuglog } from "../utils";
 
 import { BACKEND_URL } from "../config";
+import { getErrorCause } from "./utils";
 
 function defaultResolver<T, P = T>(data: T): T {
   return data;
@@ -179,7 +180,7 @@ class ApiClient {
     return this.resolveOrReject(
       this.authGet<User, {}>("/api/v1/auth/user/"),
       defaultResolver,
-      (error) => new Error(`Failed to get current user: ${error}`),
+      (error) => new Error(`Failed to get current user: ${getErrorCause(error)}`),
     );
   }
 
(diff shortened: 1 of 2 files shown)


hmpf (commented):
How would I test this manually?

podliashanyk:
> How would I test this manually?

## Selection of errors with the steps to reproduce them:

### **Incidents page**

- [ ] Dirty incident update

1. Open incidents page in 2 tabs
2. Select auto update "Interval"/"Realtime" in the first one, and "Never" in the other one
3. In the first tab, toggle either open or acked state on any incident
4. After step 3 is successful, perform the same operation on the same incident in the second tab (the one with auto update of "Never")
5. Snackbar with an error message "_Failed to close incident `<pk>` - Error: `<error description>`: {"type":`<error cause>`}_" appears at the bottom of the page

- [ ] Expiration date set to past time

1. Select any "Non-acked" incident to acknowledge
2. Write any acknowledgment message, set expiry date to any date in past
3. Submit the acknowledgment event
4. Snackbar with an error message "_Failed to post ack Error: Failed to post incident ack: {"non_field_errors":["'expiration' must be after 'event.timestamp'."]}_" appears at the bottom of the page
-----

### **Timeslots page**

- [ ] New timeslot has a name that already exists

1. Type "All the time" in a "Timeslot name*" field of a "Create new timeslot" form
2. Press "Create" in a "Create new timeslot" form
3. Snackbar with an error message "_Failed to post notificationprofile timeslot: ["Timeslot with the name 'All the time' already exists for the user `<username>`."]_" appears at the bottom of the page


- [ ] New timeslot's end time is before start time

1. Type <any non-empty string> in a "Timeslot name*" field of a "Create new timeslot" form
2. Select any start time for a new timeslot
3. Select an end time that is before the start time from step 2
4. Press "Create"
5. Snackbar with an error message "_Failed to post notificationprofile timeslot: {"time_recurrences":[{"non_field_errors":["'start' must be before 'end'."]}]}_" appears at the bottom of the page

hmpf (approved):

PR #282, opened 2021-07-01
Test/timeslots

hermar98:
This pull request includes:
- Integration test for the Timeslots-page.
- Unit tests for all the components related to the Timeslots-page.

A new dependency is added to the project, so run 'npm ci' before starting the application.

### DIFF
--- a/src/components/timeslot/index.tsx
+++ b/src/components/timeslot/index.tsx
@@ -105,21 +105,23 @@ const DEFAULT_TIMESLOT_RECURRENCE: TimeRecurrence = {
 };
 
 export type TimeslotRecurrenceComponentPropsType = {
-  pk?: TimeslotPK;
   id: number;
   recurrence: TimeRecurrence;
   onChange: (id: number, recurrence: TimeRecurrence) => void;
   onRemove: (id: number, recurrence: TimeRecurrence) => void;
   disabled?: boolean;
+
+  // Primary key of the timeslot which the recurrence belongs to. The timeslot pk is combined with the id of the recurrence to create an unique id for sub-components.
+  timeslotPk?: TimeslotPK;
 };
 
 export const TimeslotRecurrenceComponent: React.FC<TimeslotRecurrenceComponentPropsType> = ({
-  pk,
   id,
   recurrence,
   onChange,
   onRemove,
   disabled,
+  timeslotPk,
 }: TimeslotRecurrenceComponentPropsType) => {
   const style = useStyles();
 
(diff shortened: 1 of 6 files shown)


src/components/timeslot/index.tsx
```diff
@@ -105,6 +105,7 @@ const DEFAULT_TIMESLOT_RECURRENCE: TimeRecurrence = {
 };
 
 export type TimeslotRecurrenceComponentPropsType = {
+  pk?: TimeslotPK;
```
hmpf, line 108: Why do you add pk here? It's supposed to be an alias for id.

src/utils.tsx
```diff
@@ -2,6 +2,7 @@ import { useState, useEffect, Dispatch, SetStateAction } from "react";
 
 import format from "date-fns/format";
 import formatDistance from "date-fns/formatDistance";
+import moment from "moment";
```
hmpf, line 5: I think "moment" used to be a dependency but was removed later on. Why is using moment necessary again?

hermar98: The function timeOfDayFromDate(), which converts a date object to a string, didn't use a official library and converted the dates "manually". This caused some bugs with the formatting, so I had to fix it to make the tests work. I decided to use moment since I have previous experience with the library, and since the already imported library 'date-fns' gave errors.

I have now looked more closely into the error with 'date-fns', and fixed it by adding a extra if-statement to check that the date object passed to the function is valid. Will push a new commit soon were I remove moment and switch to date-fns instead.

src/components/timeslot/index.tsx
hmpf: In the backend: "pk" === "id", and is created as the recurrence is saved the first time. We only need one of them in the API for that reason, and we only need one of them in the frontend for that reason, and whatever it is called, it does not exist until first save.

src/utils.tsx
hmpf: Good, multiple explicitly imported thid-party libraries to do the same thing is a bit silly.

src/components/timeslot/index.tsx
hermar98: The two values are not equal in this case. "pk" is actually the primary key of the Timeslot which the recurrence belongs to. "id" is a value which is assigned to each recurrence in the frontend when rendered. It is used to reference a specific recurrence in the recurrence table for a timeslot (timeRecurrences[id] for example). The reason for using both is to create a unique id for each component inside a recurrence. By combining the pk of the timeslot ("pk") with the id of the recurrence ("id") we can create an unique id, which can be assigned to the components that are being rendered inside a recurrence.

Will renaming "pk" to "timeslotPk" make it easier to understand what is going on? Can also add a comment about it in the code.

hmpf: I think that might be wise, especially as there's an overhaul needed in the database schema in the backend which might change which way the dependencies go. The frontend being more self-documenting would be a smart move.

hmpf (approved):

PR #297, opened 2021-08-02
Realtime connection fallback

### LINKED ISSUE #290: Fallback to interval reloading when realtime websocket connection breaks down
### Background

The websockets-based realtime incidents updates tends to break down quite often. This may affect whether the incident list is updated at all, which would be **bad**. A very *annoying* effect is also the one that messages about lost and re-established realtime connections keep popping up at the bottom of the Argus frontend.

I've expressed this before, but we had limited resources to do anything about it:

Realtime updates are **not** critical for Argus. Refreshing on a set interval is good enough for most use cases. Therefore:

### The desired functionality

The frontend should *automatically* fall back to the interval-reloading approach if there is a problem with the websocket connection, rather than require the user to manually open the frontend settings and switch the method there.

hermar98:
The frontend will now fall back to 'interval' as auto update method if 'realtime' fails. The user will get an alert that the auto update method has changed.

I also implemented saving of the current auto update method to local storage, as I thought it was a nice thing to have.

closes #290


hmpf (changes requested):
There are some merge conflicts now.

hmpf (commented):
Typescript complains here:
```

Failed to compile.
HIDDEN/Argus-frontend/src/components/incidenttable/FilteredIncidentTable.tsx
TypeScript error in /home/hm/GIT/campus/Argus-frontend/src/components/incidenttable/FilteredIncidentTable.tsx(203,7):
No overload matches this call.
  Overload 1 of 2, '(props: { component: "div"; } & Pick<TableCellProps, "children" | "ref" | "abbr" | "slot" | "style" | "title" | "defaultChecked" | "defaultValue" | "suppressContentEditableWarning" | ... 259 more ... | "sortDirection"> & { ...; } & CommonProps<...> & Pick<...>): Element', gave the following error.
    Property 'onPageChange' is missing in type '{ rowsPerPageOptions: number[]; component: "div"; count: number; rowsPerPage: number; page: number; onChangePage: (event: unknown, newPage: number) => void; onChangeRowsPerPage: (event: ChangeEvent<HTMLInputElement>) => void; nextIconButtonProps: { ...; }; backIconButtonProps: { ...; }; SelectProps: { ...; }; }' but required in type '{ ActionsComponent?: "object" | "big" | "small" | "sub" | "sup" | "abbr" | "address" | "article" | "aside" | "b" | "bdi" | "bdo" | "blockquote" | "caption" | "cite" | "code" | "col" | ... 63 more ... | undefined; ... 14 more ...; SelectProps?: Partial<...> | undefined; }'.
  Overload 2 of 2, '(props: DefaultComponentProps<TablePaginationTypeMap<{}, ComponentType<Pick<TableCellProps, "children" | "ref" | "abbr" | "slot" | "style" | "title" | "defaultChecked" | "defaultValue" | "suppressContentEditableWarning" | ... 259 more ... | "sortDirection">>>>): Element', gave the following error.
    Type '{ rowsPerPageOptions: number[]; component: string; count: number; rowsPerPage: number; page: number; onChangePage: (event: unknown, newPage: number) => void; onChangeRowsPerPage: (event: ChangeEvent<HTMLInputElement>) => void; nextIconButtonProps: { ...; }; backIconButtonProps: { ...; }; SelectProps: { ...; }; }' is not assignable to type '(IntrinsicAttributes & Pick<TableCellProps, "children" | "ref" | "abbr" | "slot" | "style" | "title" | "defaultChecked" | "defaultValue" | ... 260 more ... | "sortDirection"> & { ...; } & CommonProps<...> & Pick<...>) | (IntrinsicAttributes & ... 3 more ... & Pick<...>)'.
      Property 'component' does not exist on type '(IntrinsicAttributes & Pick<TableCellProps, "children" | "ref" | "abbr" | "slot" | "style" | "title" | "defaultChecked" | "defaultValue" | ... 260 more ... | "sortDirection"> & { ...; } & CommonProps<...> & Pick<...>) | (IntrinsicAttributes & ... 3 more ... & Pick<...>)'.  TS2769

    201 | 
    202 |     return (
  > 203 |       <TablePagination
        |       ^
    204 |         rowsPerPageOptions={[10, 25, 50, 100]}
    205 |         component="div"
    206 |         count={count}
```

hmpf:
Ugh, I've checked out a fresh version of the repo and now gets this in other PR's as well.

hmpf:
If I change onChangePage to onPageChange in line 209 in `src/components/incidenttable/FilteredIncidentTable.tsx` it manages to build, but the resulting list of incidents is always empty :/

hermar98:
This problem is related to having the wrong version of a dependency. Running 'npm ci' instead of 'npm install' will solve the issue, as 'npm ci' installs the packages specified by package-lock.json.

hmpf (approved):

PR #306, opened 2021-08-04
Add tips-section for dev on linux

hmpf:
.. currently: how to get rid of a warning about too many watchers that
might prevent tests from running.


lunkwill42 (changes requested):
I don't usually run this on Linux, so I cannot confirm the veracity, but it is probably a useful note the day I do attempt it on Linux.

A few simple typos are all I have to complain about ;)

PR #313, opened 2021-08-11
Redesigning Notification profile page

### LINKED ISSUE #305: Make the Notification Profiles page more mobile friendly
Here are some ideas for how the Incidents page can be improved to work better on mobile devices:

- **Make notification profile boxes responsive:** Currently the notification profile boxes have more or less a fixed size. Since this size is too big for a mobile screen, the boxes will be partially rendered outside the frame. The boxes have to be made more responsive, so they work both on desktop and on mobile devices.

Screenshots of what the Notification Profiles page currently looks like on a mobile device (iPhone X):

image

hermar98:
This PR contains a total rewrite of all the components of the "Notification Profiles"-page. The main changes are:

- The page is now more mobile friendly. All components are reponsive and adapts to the actual screen size.
- New button makes it possible to add a new phone number directly from the "Notification Profiles"-page.
- "Create new profile"-button is moved to the top of the page, to make it accessible without scrolling.
- Updating the timeslot of a notification profile now actually looks like an update to the user. "Under the hood" the profile is actually deleted and a new one is created (due to limitations in the data model), but the user doesn't need to know about this.
- Required-fields are validated before a request is sent to the API, and a error message is shown if required fields are missing.
- Some minor bug fixes.

Closes #305.


hmpf (commented):
Works, but I noticed that the error-message for invalid phone numbers could be better.

The backend library used to validate numbers actually checks that it is in a valid range for the country in question, so you cannot test with any random value of the correct length. That is not the problem. The problem is that the backend insists on the country code or the number will never be considered valid.

hmpf (approved):

PR #315, opened 2021-08-11
Make Incidents-page responsive to different screen sizes

### LINKED ISSUE #299: Make the Incidents page more mobile friendly
Here are some ideas for how the Incidents page can be improved to work better on mobile devices:

- **Hide all/some of the filter settings:** Currently the Filter Toolbar covers the entire screen on a mobile device, and the user has to scroll down to even see the incidents table. I think it would be a good idea to hide some/all of the filter settings behind [an accordion](https://material-ui.com/components/accordion/). This way the user can see the incidents table without scrolling down, but also have easy access to the filter settings when they are needed.
- **Make the incidents table more mobile friendly:** Currently each row in the incidents table is so long that the user has to scroll horizontally to see all the information. This is not optimal for a mobile device. Fitting large tables onto a small mobile device is not an easy task, but [this article](https://medium.com/appnroll-publication/5-practical-solutions-to-make-responsive-data-tables-ff031c48b122) describes some practical solutions that may be used. Each solution has some pros and cons, so we have to discuss what would work best in our situation.

Screenshots of what the Incidents page currently looks like on a mobile device (iPhone X):

Skjermbilde 2021-08-03 075458

podliashanyk:
Closes #299


podliashanyk:
## Changes made:

- Add styling for small devices (f.e. iPad) 
    - Source-, Tags-, Severity- and Filter selectors are placed in accordion
    - Table header becomes sticky on scroll
    - Status and Severity columns are merged into one
    - Actions column is moved to the center of the table
Screenshot 2021-08-13 at 10 02 41

- Add styling for x-small devices (f.e. iPhone 6/7/8/X, Galaxy S5) 
    - Source-, Tags-, Severity- and Filter selectors are placed in accordion
    - Table header becomes sticky on scroll
    - Status and Severity columns are merged into one
    - Ticket url and Source columns are merged into one
    - Description and Actions are moved into accordion
    - Colourful left-side border is removed
Screenshot 2021-08-13 at 10 03 03


## Manual testing:

1. Alternative 1
   1. Go to /incidents
   2. Right click anywhere on the Incidents-page and select "Inspect"
   3. In the "Inspect" panel: click on "Toggle Device Toolbar"
   4. Select different options from "Responsive"-dropdown and observe style changes based on device size
  
2. Alternative 2
    1. Go to /incidents
    2. Manually resize browser window

src/views/incident/IncidentView.test.tsx
```diff
@@ -180,7 +180,20 @@ beforeEach(() => {
...
+      "api-version": {
+        "stable": "v1",
+        "unstable": "v2"
+      },
+      "jsonapi-schema": {
+        "stable": "/api/schema/",
+        "v1": "/api/v1/schema/",
+        "v2": "/api/v2/schema/"
+      }
+    });
```
hmpf, line 196: Wrong branch for this patch but we do need the tests, so I'll let it pass.

src/views/incident/IncidentView.test.tsx
```diff
@@ -703,6 +717,19 @@ describe('Incidents Table: reflects incident modifications', () => {
...
+          "server-version": "1.2.3",
+          "api-version": {
+            "stable": "v1",
+            "unstable": "v2"
+          },
+          "jsonapi-schema": {
+            "stable": "/api/schema/",
+            "v1": "/api/v1/schema/",
+            "v2": "/api/v2/schema/"
+          }
```
hmpf, line 732: See previous comment.

src/views/incident/IncidentView.test.tsx
```diff
@@ -768,6 +795,19 @@ describe('Incidents Table: reflects incident modifications', () => {
...
```
hmpf, line 810: Uhm, a third copy? Why?

src/views/incident/IncidentView.test.tsx
```diff
@@ -832,6 +872,19 @@ describe('Incidents Table: reflects incident modifications', () => {
...
```
hmpf, line 887: Would it be possible to move the inner object mocking the reply from the server to a named constant so that there is only one place to update when/if the API changes?

This bit:

```json
{
  "server-version": "1.2.3",
  "api-version": {
     "stable": "v1",
     "unstable": "v2"
  },
  "jsonapi-schema": {
     "stable": "/api/schema/",
     "v1": "/api/v1/schema/",
     "v2": "/api/v2/schema/"
  }
}
```

podliashanyk: Will do! This one was written in a hurry. 😅

hmpf (approved):

PR #320, opened 2021-08-16
Make modals and pop-ups responsive to different screen sizes

### LINKED ISSUE #300: Make modals, pop-ups more mobile friendly
Here are some ideas for how the Incidents Details modal/page can be improved to work better on mobile devices:

- **Make modal fill entire screen:** On a desktop computer the modal works pretty well, and gives quick access to the incident details without leaving the main incidents page. However, on a small mobile device the modal feels squished together. I think it would be better to have the modal fill the entire screen on mobile devices. A possible solution is to have it identical to the Incidents Details page, just with an X/back arrow above the header.
- **Show the most important information first:** The Incident Details modal/page contains a lot of information. On a desktop computer all the information is immediately visible to the user. However, on a mobile device the user has to scroll down to see all the information. I think it would be a good idea to put the most important information on the top of the page, so it is immediately visible to mobile users as well.
- **Hide the least important data:** The details that are less important can be hidden using [an accordion](https://material-ui.com/components/accordion/). That way the user doesn't have to scroll that much, and the modal/page will look cleaner.

Screenshots of what the Incidents Details modal and page currently looks like on a mobile device (iPhone X):

image
image

podliashanyk:
Closes #300

### DIFF
--- a/src/components/dialogs/ConfirmDialog.tsx
+++ b/src/components/dialogs/ConfirmDialog.tsx
@@ -5,8 +5,6 @@ import DialogActions from "@material-ui/core/DialogActions";
 import DialogContent from "@material-ui/core/DialogContent";
 import DialogContentText from "@material-ui/core/DialogContentText";
 import DialogTitle from "@material-ui/core/DialogTitle";
-import useMediaQuery from "@material-ui/core/useMediaQuery";
-import { useTheme } from "@material-ui/core/styles";
 
 export type ConfirmDialogPropsType = {
   title: string;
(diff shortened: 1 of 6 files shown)


podliashanyk:
## Changes made:

- Incident table toolbar is responsive to xs and sm devices
    Screenshot 2021-08-23 at 09 38 41

- Confirmation dialogs are no longer full-screen for xs and sm devices
    Screenshot 2021-08-23 at 09 38 01

- Incident detail is responsive to xs and sm devices:
    - Incident detail dialog is now full-screen
    - Background papers are removed in order to utilize more screen space
    - Operations on incidents are moved to separate component, namely "Actions"
    - Acknowledgement- and event logs are at the very bottom of the view
    Screenshot 2021-08-23 at 09 40 21
    Screenshot 2021-08-23 at 09 40 55

hmpf (approved):

Issue #175, opened 2020-12-09
Bug: Feide logout does not work


lunkwill42:
**Describe the bug**
If logging in to Argus using Feide, I cannot properly log out again. 

* If I select the "Log out" menu item, I am simply sent to the login URL. If change the URL to go back to the front page of Argus, I am still logged in as my Feide user.
* Event worse: If I log out, and then attempt to log back in using a *local* username/password combination, once I get back in to the Argus front page, Argus still thinks I am logged in as my Feide user.

**To Reproduce**
Trivially reproduced from the description above.

**Expected behavior**
When I click log out, I expect to be permanently logged out and not able to use Argus again until I log in again. I also expect to be logged in as the user whose credentials I supply, not as someone else, which is a potential security issue!

jorgenbele:
This seems to be an issue regarding how cookies are deleted on logout. When there are more than one cookie from the same domain (which happens in some circumstances apparently), it only removes one of them. I'll have to investigate some more to find out the issue here, but note that this is not reproducible in the development setup.

hmpf:
Possibly connected to #79 and #322

Issue #79, opened 2020-06-22
Call logout endpoint in backend


ddabble:
Currently, only the auth token cookie is removed when the user logs out.

PR #332, opened 2021-09-20
Fix global state for authentication

### LINKED ISSUE #322: Bug: incomplete logout when using several tabs
**Describe the bug**
When Argus is running in several tabs in the browser, logout in one of them does not trigger the logged out state in the rest of the tabs. 
API calls can be performed in all tabs, where logout was not performed explicitly. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to /incidents
2. Open /timeslots in a new tab
3. Perform logout in the /incidents tab
4. Perform any update on a timeslot in the /timeslots tab
5. See the update in the /timeslots tab being successfully performed

**Expected behavior**
Logout from Argus should trigger logout in all tabs running Argus with a given cookie.

podliashanyk:
Changes made:

- [x] Call logout endpoint
- [x] Use token that is stored in browser cookies
- [x] Update redirects
- [x] Update tests

Closes #322, #79, #175


podliashanyk:
Implemented changes:

- Logout endpoint is called
- Token from browser cookies is used in order to check for whether session is valid
- Protected route redirect (to login page) is fixed
- Implemented redirect to /incidents from login is user is already authenticated

As a result:

- login/logout is global, regardless how many Argus tabs/windows (of the same browser) user has opened. Login/logout in one of the tabs is correctly reflected in all active Argus tabs/windows (of the same browser)
- redirects correctly reflect authentication state

src/store.tsx
```diff
@@ -1,35 +0,0 @@
...
-    default:
-      return initialState;
-  }
-}
-
-export function StoreProvider(props: any) {
-  const [state, dispatch] = React.useReducer(reducer, initialState);
-  const values = { state, dispatch };
-  return <Store.Provider value={values}>{props.children}</Store.Provider>;
-}
```
hmpf, line 35: Getting rid of this was a good choice considering the naming: "store", but only of auth-relevant things.

hmpf (commented):
Seems to be working as advertised.

hmpf (approved):
Ugh, hit "Submit" to soon.

PR #334, opened 2021-09-27
Polish filter-look for small devices

### LINKED ISSUE #319: Polish: Filter-look on incidents page when at minimum screen size
The look and feel of the various inputs on the incidents-page could stand to be adjusted a little, see screen shot:

image

* The row of buttons up top are always available and cannot be hidden
* The box called "extra filters" are not actually extra, the separation between these and the button rows above is iffy
* In order to control realtime strategy and set a time range, it is necessary to first open "extra filters", then click another button. While the time range is definitely a filter, realtime strategy is not.

I think a more natural way of splitting this up is:
* Filters that can be used to control notifications (the two top rows + "extra filters")
* Time range filter
* Realtime strategy

I suggest as a step one to move time range filter into "extra filters" and mark it as not stored on the server somehow, and move the button to access realtime strategy out of the extra filters box.

In this case, it is definitely best to make some mockups.

If we implement #230 we can always hide all these the controls at minimum width behind a button, but the realtime strategy switch isn't affected by #230 at all so we'd need a solution for that regardless.

podliashanyk:
Closes #319

### DIFF
--- a/src/components/incident/IncidentFilterToolbar.css
+++ b/src/components/incident/IncidentFilterToolbar.css
@@ -19,19 +19,11 @@
 @media (max-width: 959px) {
     .lg-xl-source-selector, .lg-xl-tags-selector,
     .lg-xl-severity-selector, .lg-xl-filter-selector,
-    .lg-xl-more-settings-item {
+    .lg-xl-more-settings-item, .lg-xl-open-state-switch,
+    .lg-xl-acked-state-switch {
         display: none !important;
     }
-    div[title="Filter selector"] > div {
-        width: auto;
-        max-width: 18em;
-        flex-grow: 1;
-    }
     #filter-select {
-        padding-right: 0.3rem;
-    }
-    div[title="Filter selector"] button {
-        padding-left: 0.3rem;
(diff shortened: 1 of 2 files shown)


podliashanyk:
Suggested solution, based on feedback in #319
Screenshot 2021-09-27 at 16 27 53
 
Screenshot 2021-09-27 at 16 28 05

Any feedback is appreciated! Especially on the placement of the extra-settings-button (gears icon).

hmpf:
Without even testing: The area above the incident list is very precious and should be as small as possible. So: move the gear next to the filter box, and maybe show the name of the chosen filter if any (placeholder if no filter chosen by name).

hmpf:
(Hiding the gear inside the burger menu is also tempting.)

podliashanyk:
> Without even testing: The area above the incident list is very precious and should be as small as possible. So: move the gear next to the filter box, and maybe show the name of the chosen filter if any (placeholder if no filter chosen by name).

New suggested solution, based on the feedback:
Screenshot 2021-10-14 at 10 53 52
Screenshot 2021-10-14 at 10 54 14

Issue #303, opened 2021-08-03
Support alternative federated login mechanisms


lunkwill42:
Argus currently supports Oauth authentication, however most national federated ID providers in Europe do not (Feide is unique here). 

Most others use SAML for login. SAML is complex, and its best to keep the logic outside the application and in sidecars or in the web server - but the app needs to support some way of communicating the currently logged in user. When running in a web server, this is usually transmitted in the REMOTE_USER variable/header.

hmpf:
Depends on #302

lunkwill42:
Obsoleted by the new integrated HTMX front-end.

Issue #302, opened 2021-08-03
Make frontend and backend more independent of each other: Login process is convoluted


lunkwill42:
Currently, the frontend asks the backend to perform the Feide authentication step, and the backend needs to know the URL of the frontend to set cookies properly.

This solution is bad: There could be multiple frontends, and there are definitely multiple API clients.

Research and suggest ways to make the login process simpler, so the backend won't need have specific knowledge of the frontend.

hmpf:
The real problem I guess is that we use a cookie for this, and that the frontend knows about how the backend talks to dataporten. I think the ideal solution would be the frontend is also a user and has a dedicated login endpoint, and sends its own location + token. The backend does the login. The backend knows the difference between a dataporten login and a token login, so after a dataporten login it sends the token directly to the frontend, just like dataporten sends data directly to the backend. The backend does not set the cookie and doesn't care what the frontend does.

For safety, we could have that the frontend needs to be known to the backend. Currently, we hardcode this, and are limited to a single frontend per backend. If a specific frontend is a user, then we can use tokens as usual. We could also have a whitelist in the database that contains the domains that are allowed to access the dataporten login endpoint, but then we run into the problem of review-sites with auto-generated domain-names again.

Hey, we could even have that the frontend has a user per login mechanism.

hmpf:
Here's how to let the frontend do the login with OAuth2:

https://www.toptal.com/django/integrate-oauth-2-into-django-drf-back-end

lunkwill42:
This has become obsoleted by the integrated HTMX UI in Argus. Closing.

PR #341, opened 2021-11-25
Fix feide logout

### LINKED ISSUE #340: Bug: Cannot log in after logging out when using dataporten/feide
Something odd happens when I log out of argus (frontend 1.5.2). 
I don't get the login-window at all, and can no longer log in. When not using containers (for instance via safe mode or private window) it works fine.

The load goes through the roof and I need to restart firefox.

This has been verified to happen on Firefox (91.0.2) both with and without containers, as well as Vivaldi 4.1.x. A workaround to be able to log back in is to manually get rid of the cookie for the domain.

See video:

https://user-images.githubusercontent.com/11956096/139822193-5151b362-f3d7-4d75-b79c-e93a7232f1d9.mp4

podliashanyk:
Fixes #340.

**Cause of the bug:**
Login via dataporten  sent responses with a "Set-Cookie" header. Backend forwards it and frontend sets cookie sent from dataporten. Frontend sets cookies from dataporten as a subdomain cookie and will not remove it, unless that specific subdomain is provided on removal request. Subdomain format is ".COOKIE_DOMAIN". Presence of "token" in cookies is equal to user being considered logged on the backend, and is a main factor in deciding on whether to call either token-auth, or auth endpoint. 

**Introduces changes:**

- requires a new env variable (REACT_APP_COOKIE_DOMAIN) in frontend config. This one is **necessary** for removal of token which is set during dataporten-auth, and is a nice-to-have for other auth methods. This is due to the fact that removal of token set during dataporten auth, domain has to be provided. Domain could be extracted from the url, but is not optimal. Cookie domain can be set in the api settings, and can theoretically be different than api url (see domain values for cookies in the review stage in argus-run).
- cookies are properly removed on logout, regardless of auth method
- post-logout cleanup is improved

**Functionality:**

- User can have several active tabs/windows of Argus. Logout in one of them is reflect in all active tabs. Restrictions apply:
      - Tabs/windows can not be of different browsers, nor same browser but incognito mode. This is due to backend not supporting several sessions of the same user. 
- Login event in a new browser, while there is an active session in another browser, will work but will also trigger logout in another browser. This does not apply if login method is different in different browsers (f.e. userpass in one, and dataporten in another).

**Manual testing suggestion:**

1. Log in via userpass/dataporten.
2. Open f.e. /incidents in another browser
3. Open f.e /incidents in a new window of the same browser
4. Log out in one of the active tabs/windows
5. Observe, one of the following (depending on auto-update method):
       - logout is triggered immediately in all other active tabs/windows
       - logout is triggered as soon as user interacts with the app in the rest of the tabs/windows (no auth-restricted operations can be performed)
6. Observe redirect to /login
7. Try go to f.e. /incidents without performing login
8. Observe immediate redirect back to /login


hmpf:
I played around (with Firefox) on the review you set up. If one repeatedly logs in and out, in the beginning there are two token cookies, but after a while it is only one of them.

src/auth.tsx
```diff
@@ -11,27 +11,40 @@ class Auth {
     this.authenticated = false;
   }
 
+  // feideLogin(token: string, callback?: () => void) {
+  //   this.authenticated = true;
+  //   this._token = token;
+  //   if (callback) callback();
+  // }
+
```
hmpf, line 19: Why keep these?

hmpf (commented):
Please document what should be set in REACT_APP_COOKIE_DOMAIN. Is it the same value as ARGUS_COOKIE_DOMAIN in the backend?

Also, I would love to know what the actual problem was, considering there are issues on supporting other login methods, and making the backend more independent from the frontend. (The frontend must obviously know where the backend is running, but the backend shouldn't ever need to know where the frontend is running.) How does login happen?

lunkwill42 (commented):
> - logout is triggered immediately in all other active tabs/windows

I did test this on Safari. Logout in the other tabs was not triggered *immediately*, there was a noticeable delay. But, I **was** logged out :)

I was also logged out in the other browser I was using.

The latter makes no sense, but I guess that is just the consequence of how the delete-token-on-logout system works. 

After this fix is merged, we still need to think about how logins *actually* should interact between the UI and the backend (i.e. the backend is just an API, it shouldn't even need to set cookies in your browser for you to use the frontend, the frontend just needs to get a token that is valid against the API).

lunkwill42:
> Please document what should be set in REACT_APP_COOKIE_DOMAIN. Is it the same value as ARGUS_COOKIE_DOMAIN in the backend?

Likely, yes. In 99% of cases, I would assume `REACT_APP_COOKIE_DOMAIN `equals the backend URL's domain name, which is already known. I think the frontend should probably default to that value if not configured explicitly (most user's won't need to care).

podliashanyk:
> I played around (with Firefox) on the review you set up. If one repeatedly logs in and out, in the beginning there are two token cookies, but after a while it is only one of them.
> 
> On Vivaldi however, this doesn't seem to happen. Both cookies are created, and both cookies are removed each time. Weird.

Could not reproduce. I tried repeatedly log in and out in Firefox for over 20 times, but always got 2 cookies. Sometimes only one was displayed by default, but as soon as clicked on "Cookies" in inspector, both "token"-cookies were there each and every time.

src/auth.tsx
podliashanyk: Removed in d554b8d

podliashanyk:
> Also, I would love to know what the actual problem was, considering there are issues on supporting other login methods (#303), and making the backend more independent from the frontend (#302). (The frontend must obviously know where the backend is running, but the backend shouldn't ever need to know where the frontend is running.) 

The actual problem was that login via dataporten  sent responses with a "Set-Cookie" header. Backend forwards it and frontend sets cookie sent from dataporten. Frontend sets cookies from dataporten as a subdomain cookie and will not remove it, unless that specific subdomain is provided on removal request. Subdomain format is ".COOKIE_DOMAIN".

podliashanyk:
> > * logout is triggered immediately in all other active tabs/windows
> 
> I did test this on Safari. Logout in the other tabs was not triggered _immediately_, there was a noticeable delay. But, I **was** logged out :)
> 
The dalay I assume was approximately equal to auto-refresh interval? Logout in other tabs/windows (the ones where logout was not performed originally) is triggered on backend's endpoint call/page reload. I expect to fix this restriction and make logout immediate _everywhere_ when I refactor global state. Current implementation of global state does not support this. 

> I was also logged out in the other browser I was using.
> 
> The latter makes no sense, but I guess that is just the consequence of how the delete-token-on-logout system works.
> 
That is due to backend's workings on login event. Calling /login _from different browsers_ on backend will either set the same token that is already present in some active session, _or_ will renew the token, devalidating the one that is already in use. I have myself not figured the pattern on whichever one will occur.

podliashanyk:
> After this fix is merged, we still need to think about how logins _actually_ should interact between the UI and the backend (i.e. the backend is just an API, it shouldn't even need to set cookies in your browser for you to use the frontend, the frontend just needs to get a token that is valid against the API).
> Also, I would love to know what the actual problem was, considering there are issues on supporting other login methods (#303), and making the backend more independent from the frontend (#302). (The frontend must obviously know where the backend is running, but the backend shouldn't ever need to know where the frontend is running.) How does login happen?

I will make a consolidated issue regarding auth! :)

podliashanyk:
> > Please document what should be set in REACT_APP_COOKIE_DOMAIN. Is it the same value as ARGUS_COOKIE_DOMAIN in the backend?

@hmpf do you have any suggestions on where such documentation belongs?

podliashanyk:
> Likely, yes. In 99% of cases, I would assume `REACT_APP_COOKIE_DOMAIN `equals the backend URL's domain name, which is already known. I think the frontend should probably default to that value if not configured explicitly (most user's won't need to care).

Fixed!

hmpf:
> > > Please document what should be set in REACT_APP_COOKIE_DOMAIN. Is it the same value as ARGUS_COOKIE_DOMAIN in the backend?
> 
> @hmpf do you have any suggestions on where such documentation belongs?

The only place we have for frontend docs right now is the README.md, so that's where it needs to go. If it is an envvar it can go with the other envvars but if not we'll need a new section about the config.tsx-file.

lunkwill42:
> The dalay I assume was approximately equal to auto-refresh interval?

That was how I perceived it, indeed.

lunkwill42 (approved):
I'm giving this a :+1:

We want it merged ASAP, but will add a separate issue for fixing the docs

Issue #336, opened 2021-10-01
Bug: "Open State = BOTH" and "Acked = BOTH" combined with filters gives fishy incident list


vidarstokke:
**Describe the bug**
When I use filters (either Tags or Sources) and combine this with "Open State = BOTH" and "Acked = BOTH", there is something fishy going on in the incident list. If I also have "Rows per page: 100", the frontend will not display 100 incidents that match the filter, but a lot less. To get more incidents that match the filter, I need to go to the next page. 

**To Reproduce**
1. Select Open State = BOTH
2. Select Acked = BOTH
3. Select Tags = "kundetjeneste=Campus_CNaaS" and "organization=uninett". 
4. Select Max level = 5
5. Select Timeframe = No timeframe

This will currently give a list of 6 items. But at the bottom line it says "1-100 of more than 100". You will have to go to next page to get the next items matching the filters. 

**Expected behavior**
I would like to see 100 incidents matching the filters and not only 6. 

**Screenshots**
![image](https://user-images.githubusercontent.com/73838355/135587604-17d05b9b-2b47-4742-bde9-97cb1429906e.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version 93.0.4577.82

podliashanyk:
This issue is related to #324

lunkwill42:
Closed by #346

PR #346, opened 2021-12-09
Fix incidents table filtering

### LINKED ISSUE #324: Bug: Tags filter only hides filtered incidents, counter messed up
**Describe the bug**
Filtering incidents using the Tags filter hides the filtered incidents, but they still occupy space on the page.
So if 6 incidents on a page are filtered out using tags, a page can state that it contains 10 incidents, but only show the 4 that are not filtered out.

**To Reproduce**
1. Go to https://argus.uninett.no/
2. Set tags to something
3. Browse through the pages and observe the number of entries per page vs stated rows-per-page

**Expected behavior**
The pages should fill up with incidents matching the filter. 
The incidents that are filtered out should not occupy space on the page, they should instead be replaced with the next incidents that match the filter.

**Screenshots**
Screenshot with no Tags used. 10 incidents are shown in the page (Rows per page=10). Two of these entries have the tag room=ifi2.

not-tags

Screenshot with tags used. Only  the two incidents from the first screenshot that match the tags filter is included on the first page, even though it states that it shows 10 entries.
tags

The second page shows 7 entries, showing that the 2 on the first page are not the only ones matching the filter

image

**Desktop:**
 - OS: Ubuntu 20.04.2
 - Browser Firefox 
 - Version 91.0

podliashanyk:
**Changes made:**

- Tags and Sources are moved to FilterContent. This is both mirroring how backend recognizes filters, as well as facilitates state's recognition of _any_ changes to the filter params. 
- State is now reactive to _any_  valid updates to the sources and tags inputs, and sends a request to the backend on _any_  valid update. Previously, updates to tags and/or sources in the filter toolbar were not triggering requests to the backend, and incidents not matching those updates were "removed" from the table on frontend (but they still occupied space).
- Invalid source- and tag formats provided by the user are silently ignored (they are removed from the input fields and do not trigger table update). Previously, the invalid formats were added to the input field and were triggering the update.

Closes #324, #336

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -395,19 +395,16 @@ class ApiClient {
       if (SHOW_SEVERITY_LEVELS && filter.filter.maxlevel !== undefined) {
         params.push(`level__lte=${filter.filter.maxlevel}`);
       }
-      if (filter.sourceSystemIds !== undefined) {
-        params.push(`source__id__in=${filter.sourceSystemIds.join(",")}`);
+      if (filter.filter.sourceSystemIds !== undefined) {
+        params.push(`source__id__in=${filter.filter.sourceSystemIds.join(",")}`);
       }
       if (timeframeStart) {
         params.push(`start_time__gte=${formatTimestamp(timeframeStart)}`);
       }
-      // if (filter.sourceSystemNames !== undefined) {
-      //   params.push(`source__name__in=${filter.sourceSystemNames.join(",")}`);
-      // }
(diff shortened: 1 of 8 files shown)


lunkwill42 (approved):
As always, not particularly skilled at reviewing the code changes, but this sure does seem to do the filtering trick, so all in all, I'd say excellent work! :)

On a side-note: CodeCov did not pick up this PR, for some reason, while it did pick up the others. Weird...

PR #367, opened 2022-04-25
Add release notes and changelogs

stveit:
Fixes #367

Adds release notes and changelog files.

They are pretty much identical rn as there isnt alot to write about

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -1,0 +1,4 @@
+# Release Notes
+This file documents changes to Argus-frontend that are important for the users to know.
+
+## [Unreleased]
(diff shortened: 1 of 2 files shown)


NOTES.md
```diff
@@ -0,0 +1,7 @@
+# Release Notes
+This file documents changes to Argus-frontend that are relevant for the users to know.
+
+## [Unreleased]
+
+### Fixed
+- Fix ARGUS logo clipping part of the name.
```
lunkwill42, line 7: I don't see this as a very important point for end users to know.

I still view the releases notes mostly as a "oh, btw, you need to do some manual action or change some config or change your workflow before/during/after an upgrade to this version")

PR #373, opened 2022-06-07
Fix inconsistent labeling

### LINKED ISSUE #333: Bug: Inconsistent labeling, unacked vs. non-acked
**Describe the bug**
There is a small, but annoying inconsistency in labeling on the Incident list:

The filter buttons for an Incident's Acked status have a button that reads *"UNACKED"*, while in the incidents ilst, an unacked incident is labeled with *"Non-acked"*.

**Expected behavior**

Both labels should read *"Unacked"*, in my opinion.

**Screenshots**

bilde

vs

bilde

**Versions**
This was observed on version 1.5.1 of Argus-frontend

podliashanyk:
Chip elements are updated to match the button in the filter toolbar:
Screenshot 2022-06-07 at 10 21 04

Unsure whether labelling in the detailed view should also be updated from "Unacknowledged" to "Unacked"..?
Screenshot 2022-06-07 at 10 20 34

Closes #333

### DIFF
--- a/src/components/incident/Chips.tsx
+++ b/src/components/incident/Chips.tsx
@@ -96,7 +96,7 @@ export const AckedItem: React.FC<AckedItemPropsType> = ({ acked, expiration, sma
 
   const expiryDate = expiration && new Date(expiration);
   if (small) {
-    return <Chip size="small" className={className} label={acked ? "Acked" : "Non-acked"} />;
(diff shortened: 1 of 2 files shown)


lunkwill42:
> Unsure whether labelling in the detailed view should also be updated from "Unacknowledged" to "Unacked"..?

I think it's fine to keep it slightly more verbose in the details page :)

lunkwill42 (approved):

PR #376, opened 2022-06-08
Make docker compose build more flexible

lunkwill42:
Mostly this is about the first commit:

Using `${UID}` as the `user` value in `docker-compose.yml` causes problems on macOS (where the mapping/privilege issues don't apply anyway). However, it seems setting it to something other than the image default of `node` might cause problems for the latest node images from Dockerhub (npm just exits with code `243`, producing no output or logs to indicate what went wrong).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,9 @@ services:
       - "8080:8080"
     volumes:
       - .:/app:Z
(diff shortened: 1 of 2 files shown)


podliashanyk (approved):
Works well!

PR #377, opened 2022-06-08
Enable the SMS media plugin in docker-compose env

lunkwill42:
This introduces a custom settings file for the API backend service container, so some of the more complex settings of the API server can be manipulated while testing/developing the frontend.

The default settings file used will enable the SMS-as-email plugin, so that developers will be able to work with dumps of production data that contain profiles with phone numbers in them (the API server will crash if there are SMS-enabled profiles present in the database when no SMS plugin is enabled in the settings).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
       - redis
     environment:
       - TIME_ZONE=Europe/Oslo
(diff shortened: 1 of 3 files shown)


podliashanyk (approved):
Great job! 🙌🏻

PR #378, opened 2022-06-09
Display seconds in all timestamps of the detailed incident view

### LINKED ISSUE #370: Add seconds to all timestamps in incident detailed view
This is a request from the Service Center. This will make their workflow for ticket creation much easier.

podliashanyk:
### Changes made:
Timestamps in the _Related events_ feed display seconds now.
Screenshot 2022-06-09 at 15 43 28

Closes #370


stveit (approved):
Works for me 👍

PR #380, opened 2022-06-09
Display incident end time

### LINKED ISSUE #379: Display end time in detailed incident view
This is a request from the Service Center.  _End time_ is useful when creating tickets. 

_End time_ can be placed between _Start time_ and _Duration_ 
Screenshot 2022-06-09 at 15 49 07

**Note:** _End time_ should come in addition to _Duration_, and not replace it. This is due to some ticket management systems relying on duration of the incident.

podliashanyk:
### Changes made:
Added _End time_ in the detailed incident view. This applies only to closed (resolved) incidents, as end time for the open and stateless incidents is irrelevant. 

Closes #379


src/components/incident/test/IncidentDetails.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+        expect(within(primaryDetailsContainer).getByText(/primary details/i)).toBeInTheDocument() // title
+        expect(within(primaryDetailsContainer).getByRole('list')).toBeInTheDocument()
+        expect(within(primaryDetailsContainer).getAllByRole('listitem').length).toBeGreaterThanOrEqual(6)
+
+        // End time item is displayed
+        const endTimeItem = within(primaryDetailsContainer).getByTitle(/end-time/i)
+        expect(endTimeItem).toBeInTheDocument()
+        if (resolvedIncidentMock.end_time) {
+            expect(within(endTimeItem).getByText(resolvedIncidentMock.end_time)).toBeInTheDocument()
+        }
```
stveit, line 107: why if here? resolvedIncidentMock is defined to have an end_time

src/components/incident/test/IncidentDetails.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+        // Duration item is displayed
+        const durationItem = within(primaryDetailsContainer).getByTitle(/duration/i)
+        expect(durationItem).toBeInTheDocument()
```
stveit, line 111: above you check if the end_time value itself is rendered. Here you dont check if the duration value is rendered, you only check if there is something within the container with title matching `/duration/i`

src/components/incident/test/IncidentDetails.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+        // End time item is not displayed
+        const endTimeItem = within(primaryDetailsContainer).queryByTitle(/end-time/i)
+        expect(endTimeItem).toBeNull()
+        expect(endTimeItem).not.toBeInTheDocument()
+
+        // Duration item is displayed
+        const durationItem = within(primaryDetailsContainer).getByTitle(/duration/i)
+        expect(durationItem).toBeInTheDocument()
+    });
```
stveit, line 137: same here, you dont actually check if the duration value is shown

stveit (commented):
think i found a couple things that can be improved

src/components/incident/test/IncidentDetails.test.tsx
podliashanyk: This if is used here because of type definition of _Incident_. In _Incident_ interface  _ent_time_ is either string or undefined. We don't have defined subtypes for _ResolvedIncident_ where _end_time_ is strictly a string. So at this point in test TS is not aware that _end_time_ in this object is defined. 
If I remove if, test will pass but linter will complain. I will replace this if with the directive to the linter to ignore the next line, since here as you say _end_time_ will always be defined.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ This file documents all changes to Argus-frontend. This file is primarily meant
...
 - Fix ARGUS logo clipping part of the name.
 
+- End time to the incident detailed view (for closed incidents only)
+
 ### Changed
 - Logo and favicon
 - Add seconds to timestamps in elements of the event feed in detailed incident view
+
+### Added
+- End time to the incident detailed view (for closed incidents only)
```
stveit, line 16: This should probably only be under the `ADDED` section.

src/components/incident/test/IncidentDetails.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+        expect(endTimeItem).toBeNull()
+        expect(endTimeItem).not.toBeInTheDocument()
+
+        // Duration item is not displayed
+        const durationItem = within(primaryDetailsContainer).queryByTitle(/duration/i)
+        expect(durationItem).toBeNull()
+        expect(durationItem).not.toBeInTheDocument()
+    });
+});
+
```
stveit, line 181: And I suppose if we're going to be super nitpicky remove the extra newline and only have 1 :P

stveit (commented):
nitty picky

CHANGELOG.md
podliashanyk: But it is already only under `ADDED` section. 😅

```
### Changed
- Add seconds to timestamps in elements of the event feed in detailed incident view
```
is there from a different PR, if you meant to remove that one

src/components/incident/test/IncidentDetails.test.tsx
podliashanyk: That one was litt of an oopsie 😂

stveit (approved):
👍

PR #383, opened 2022-06-16
Add filtering after stateless incidents via button

podliashanyk:
Option "Open": only open and stateful incidents; Option "Closed": only closed and stateful incidents; Option "Stateless": open/closed (for legacy incidents that may be present in db) but stateless incidents; Option "All": open/closed and stateless/stateful incidents. "Open state" Button switch is renamed to "Progress state".

Closes Uninett/Argus#1438

### DIFF
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -10,7 +10,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: SonarQube Scan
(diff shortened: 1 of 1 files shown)


podliashanyk:
After conversation with @lunkwill42:
It seems like the need for frontend-modifications regarding _stateless incidents_ are unnecessary. 
What Service Center wants, can be achieved by setting _Severity_ to lower than 5. It seems there was a misunderstanding regarding what _stateless incidents_ actually represent.

PR #402, opened 2022-08-31
Remove commented out code

podliashanyk:
Remove commented out code as suggested in _Argus code audit_ [July 2022] by GÉANT.

Closes #394

PR #403, opened 2022-08-31
Remove todo comments, or implement changes as suggested in comments

### LINKED ISSUE #393: Remove todo-comments in code
**Todays situation:**
In Argus-frontend code base, there are code bits/functions that are marked with comments suggesting that some code is not finished, or need polish/refactoring. 

**Suggested solution:**
Comments pointing out not-finished code are not many. Additionally, upon further inspection, it is apparent that some of those comments are not relevant anymore.
Therefore, the suggested solution is to:

- **either** finish/polish the code bit as suggested in the comments
- **or** remove the comment if changes are not relevant anymore

**Credit:**
This issue was recommended by GÉANT in _Argus code audit_ [July 2022]. 

**Additional info:**
All files (including line numbers) where this issue is present are listed in _Argus code audit_.

podliashanyk:
Go through unfinished code as suggested in _Argus code audit_ [July 2022] by GÉANT.

Closes #393

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -308,20 +308,6 @@ class ApiClient {
       });
   }
 
-  // Incident
-  // WIP TODO untested
-  public putIncident(incident: Incident): Promise<Incident> {
-    if (Date.now() % 2 === 0) {
-      return Promise.reject(new Error(`Failed to put incident`));
-    }
-    return Promise.resolve(incident);
-    // return this.resolveOrReject(
-    //   this.authPut<Incident, Incident>(`/api/v1/incidents/${incident.pk}`, incident),
-    //   defaultResolver,
-    //   (error) => new Error(`Failed to put incident: ${error}`),
-    // );
-  }
-
   public postIncidentReopenEvent(pk: number): Promise<Event> {
     return this.resolveOrReject(
       this.authPost<Event, EventWithoutDescriptionBody>(`/api/v1/incidents/${pk}/events/`, { type: EventType.REOPEN }),
@@ -438,8 +424,6 @@ class ApiClient {
     );
   }
 
-  // NOTE: This won't actually get all incidents anymore because of the pagination
-  // TODO: Remove all use of this method.
   public getAllIncidentsFiltered(filter: Omit<Filter, "pk" | "name">): Promise<Incident[]> {
     return this.getPaginatedIncidentsFiltered(filter, null).then(paginationResponseResolver);
   }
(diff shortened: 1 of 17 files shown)


src/services/RealtimeService.ts
```diff
@@ -212,7 +212,7 @@ export class RealtimeService {
     this.ws.onerror = (e: any) => {
       console.error(`[RealtimeService ${this.id}] got error on websocket client, closing: ${e.message}`);
       if (this.ws) this.ws.close();
-      // this.setState("closed"); // TODO: should this be set?
+      this.setState("closed"); 
     };
```
stveit, line 216: Is this the same change as in PR #402 ?

src/views/incident/IncidentDetailView.tsx
```diff
@@ -29,9 +29,7 @@ export const IncidentDetailsView: React.FC<IncidentDetailsViewPropsType> = () =>
         displayAlert(`Failed to get incident with pk=${pk}`, "error");
         setIncident(null);
       });
-    // FIXME:
-    // eslint-disable-next-line
-  }, [pk]);
+  }, [pk, displayAlert]);
```
stveit, line 32: Where does this `displayAlert` come from? does not seem to be relevant to any comments?

src/api/client.ts
```diff
@@ -438,8 +424,6 @@ class ApiClient {
     );
   }
 
-  // NOTE: This won't actually get all incidents anymore because of the pagination
```
stveit, line 441: seems like a reasonable comment tbh, name is misleading and the comment corrects it (ASSUMING that this comment is even true, I have no idea. if its not true, then ofc remove it). Best would be to make the name less misleading, but since the next comment is "// TODO: Remove all use of this method.", its probably not worth putting energy into renaming if its gonna be removed soon.

stveit (commented):
comments based on ignorance of the codebase

src/services/RealtimeService.ts
podliashanyk: Yes

src/views/incident/IncidentDetailView.tsx
podliashanyk: That's what `FIXME` refers to. Without _displayAlert_ in dependencies, the program will not compile when `// eslint-disable-next-line` is removed

src/api/client.ts
podliashanyk: This comment is outdated

src/views/incident/IncidentDetailView.tsx
stveit: aaah okay, its some react shenanigans with useEffect

stveit (approved):
guess i dont have anything more to say other than 👍

PR #407, opened 2022-09-07
Update logo

### LINKED ISSUE #405: Replace current Argus logo with the newly designed one
The frontend UI should change its logo and favicons ASAP, to the new logos uploaded to #385 .

Not sure if the current logo is actually in use in the backend anywhere, but if it is, a separate issue should be opened for that codebase.

**EDIT:**
Requirements we discussed during the logo selection meeting:
* The logo assets containing the "tv screen" background element aren't really wanted, just the plain versions containing the "eye" and "Argus" text elements.
* The site favicon shouldn't be too complicated; it was suggested that the favicon remain just the "eye" element, without the "argus" text below it.

podliashanyk:
Closes #405


podliashanyk:
> Quite to the point - but: I should have written down the requirements we discussed during the IRL logo selection meeting from last week:
> 
> * We don't want the "tv screen" version of the logo assets, just the plain combo of the "eye" and the "Argus" text element.
We don't have the version of logo you mention in the bundle in white colour. I tried editing the colour myself, but edges look weird when I do so (proper editors on Mac or online require payed subscriptions 😅). The other alternative is to change the Argus-header colour.

> * We also suggested using just the eye, without the text element, for the site favicon (I don't think the asset bundle has a file with only the eye, so a new SVG with the text element removed should be produced).
We have just the eye in the bundle. I just thought that it looks slightly weird as a favicon to be honest. I can change it to eye only.

lunkwill42 (approved):
Beautiful 👍 😄

PR #411, opened 2022-09-14
Flip incident events order

### LINKED ISSUE #392: Flip the order of events in detailed incident view
This is a request from the Service Center.
They would prefer an order where oldest events and acknowledgements are displayed first in the _Event feed_ and in the _Acknowledgements feed_.

podliashanyk:
Closes #392


stveit (approved):
short and sweet 👍

PR #412, opened 2022-09-14
Add scrollbar to event and ack feeds in detailed incident view

podliashanyk:
Closes #388


stveit:
some visual feedback. for context it looks like this for me: 
image

- It feels a bit weird that the scrollbox doesnt exend as far down as the info on the left. Basically creates wasted space below the scrollbox. If all the space was used you could see more events at a time and in some specific instances you would avoid having to scroll at all.
- horizontal scrollbar being visible when I'm pretty sure you would never have to horizontal scroll also feels weird.
- `CREATE ACKNOWLEDGEMENT` button should probably have margin or something above it if the horizontal scroll bar is to stay, right now they are pretty much touching each other
- I am not sure what the conventional way is, so stop me if there is some standard I dont know about, but we could consider hiding the scrollbar if its not needed. In the Related Events column in my example there is not enough events to need scrolling, but the scrollbars are still visible

johannaengland (approved):
Looks very nice!

stveit (approved):
I found a small thing where when there are enough acks for them to be scrollable, the `CREATE ACKNOWLEDGEMENT` button touches the text after scrolling partway down, like so:
image
But its not a big deal so I'll approve and you can decide if its worth the time

PR #415, opened 2022-09-16
Make timeslots reusable

### LINKED ISSUE #213: Not possible to use a timeslot in multiple notification profiles
**Describe the bug**
I need to have the ability to reuse a timeslot in different notification profiles. This because I for instance want incidents hitting different filters to be sent to different media. 

For instance I would like to have incidents during "Working hours!" with filter "A" sent to media Email. And at the same time I would like to have incidents during "Working hours!" with filter B sent do media SMS. 

This is not possible now. 

**To Reproduce**
1. Go to 'Profiles'
2. Create a new profile with timeslot "Working hours!" with a random filter and media (for instance SMS).
3. Click save on that profile.
4. Create another profile. It will then not be possible to choose "Working hours!" timeslot because it is already in use. 

**Expected behavior**
It should be possible to use the same timeslot more than once.

podliashanyk:
Closes #213, #445


podliashanyk:
The changes in logic are to be considered completed at this point, but there are still changes to be made after #416 is closed:

- API calls in _client.ts_ will need to be updated after the notion of _destinations_ is introduced to the frontend.
- Fix failing tests after API v2 endpoints are fixed.
- Write new tests after API v2 endpoints are fixed.

hmpf:
It seems to work fine. Now for a quick code-dip.

hmpf (approved):
Yoooo so much deleted! Removing unnecessary code is the bestest change.

PR #425, opened 2022-10-04
Update changelog with changes implemented before changelog was added

podliashanyk:


lunkwill42 (commented):
Looks OK, but is there no record of which version numbers these changes were released in?

hmpf (approved):

PR #439, opened 2022-10-10
Make phonenumber component prettier

### LINKED ISSUE #434: Update phone number UI
The phone number components are a lil bit janky. This is how it looks like for me now:
image
the `SAVE` and `DELETE` buttons are very close to each other, some margin would probably make it look a little nicer.
The buttons are also not vertically centered, i think that would make it look better as well.

**Desktop (please complete the following information):**
 - OS: Ubuntu Linux
 - Browser: Firefox Version 105.0

stveit:
Fixes #434 
Updates  css for the phonenumber component.
Vertically aligns buttons, adds margins and evenly spaces the buttons out inside the paper.
Removes CSS that either wasnt used or didnt do anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,9 @@ This file documents all changes to Argus-frontend. This file is primarily meant 
 
 ### Fixed
 - Fix ARGUS logo clipping part of the name.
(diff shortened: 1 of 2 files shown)


hmpf (changes requested):
Works, but fix the conflict in the CHANGELOG

podliashanyk (approved):
Very nice!

hmpf (approved):

PR #441, opened 2022-10-10
Fix enter button not saving/updating telephone number

stveit:
Previosuly the button worked by having an OnClick event. But at the same time the form still seemed to have a submit event
that triggered when you press enter inside the form, and the default behaviour is to refresh the page.
This fix puts the OnClick logicinside a OnSubmit handler instead, while disabling the default submit behaviour (stops the page from rereshing). This makes it so
pressing enter and clicking the save/create button behaves the same way- They should both behave the same as the old behaviour for clicking the button.

Fixes #433


hmpf (changes requested):
Works, fix the CHANGELOG and we are go.

Issue #423, opened 2022-09-28
Fix helper text in dialogs for bulk-operations


podliashanyk:
Suggested by @johannaengland in https://github.com/Uninett/Argus-frontend/pull/410#discussion_r982197715.

Dialogs for bulk-close and bulk-ack have helper text with _incident_ in singular. It should be changes to plural since it is a bulk-operation. 

Suggested helper for bulk-close: "Write a message describing why the incidents were manually closed".
Suggested helper for bulk-close: "Write a message describing why the incidents were acknowledged".

- [x] Update helper text
- [ ] Update screenshots in the user manual

podliashanyk:
The remaining task is moved to a new issue #465

PR #444, opened 2022-10-11
Use plural words when bulk operating multiple incidents

stveit:
Fixes parts of #423 
also removes some trailing whitespace

Uses spread operator to only override the text if multiple incidents are selected. Is this a clean way to do it?

### DIFF
--- a/src/components/incident/CreateAckSignOffAction.tsx
+++ b/src/components/incident/CreateAckSignOffAction.tsx
@@ -52,7 +52,7 @@ const CreateAck: React.FC<CreateAckPropsType> = ({ onSubmitAck, signOffActionPro
 
   const signOffActionPluralProps = {
     dialogContentText: "Write a message describing why the incidents were acknowledged",
-    question: "Are you sure you want to acknowledge these incident?",
+    question: "Are you sure you want to acknowledge these incidents?",
   }
 
   return (
--- a/src/components/incident/CreateAckSignOffAction.tsx
+++ b/src/components/incident/CreateAckSignOffAction.tsx
@@ -38,7 +38,7 @@ const CreateAck: React.FC<CreateAckPropsType> = ({ onSubmitAck, signOffActionPro
   return (
     <SignOffAction
       dialogTitle="Submit acknowledment"
-      dialogContentText="Write a message describing why this incident was acknowledged "
+      dialogContentText="Write a message describing why this incident was acknowledged"
       dialogSubmitText="Submit"
       dialogCancelText="Cancel"
       dialogButtonText="Create acknowledegment"
--- a/src/components/incident/CreateAckSignOffAction.tsx
+++ b/src/components/incident/CreateAckSignOffAction.tsx
@@ -15,9 +15,10 @@ import SignOffAction, { SignOffActionPropsType } from "./SignOffAction";
 type CreateAckPropsType = {
   onSubmitAck: (ack: AcknowledgementBody) => void;
   signOffActionProps?: Partial<SignOffActionPropsType>;
+  isBulk: boolean;
 };
 
-const CreateAck: React.FC<CreateAckPropsType> = ({ onSubmitAck, signOffActionProps = {} }: CreateAckPropsType) => {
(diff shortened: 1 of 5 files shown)


podliashanyk (changes requested):
Good grasp on spread operator! :) 
But I would suggest refactoring of the _CreateAckSignOffAction.tsx_ and _ManualCloseSignOffAction.tsx_ instead, and making them aware of whether they are being used for bulk operation or not. This is because there are other props that should also be updated for the bulk components:

- in CreateAck - _question_
- in ManualClose - _dialogTitle_, _question_, _title_ (maybe)

So the suggested solution is to add a new prop to CreateAck and ManualClose (something like isBulk) and update the components from there.

hmpf (approved):
Will merge when @podliashanyk is happy.

src/components/incident/CreateAckSignOffAction.tsx
```diff
@@ -35,19 +36,29 @@ const CreateAck: React.FC<CreateAckPropsType> = ({ onSubmitAck, signOffActionPro
...
+    isDialogInputRequired:false,
+    dialogInputType: "text",
+    title: "Submit acknowledment",
+    question: "Are you sure you want to acknowledge this incident?",
+    onSubmit: handleSubmit,
+  }
+
+  const signOffActionPluralProps = {
+    dialogContentText: "Write a message describing why the incidents were acknowledged",
+    question: "Are you sure you want to acknowledge these incident?",
```
podliashanyk, line 55:
```suggestion
    question: "Are you sure you want to acknowledge these incidents?",
```

src/components/incident/ManualCloseSignOffAction.tsx
```diff
@@ -29,23 +30,35 @@ const ManualClose: React.FC<ManualClosePropsType> = ({
...
+    dialogInputType: "text",
+    title: "Manually close incident",
+    question: "Are you sure you want to close this incident?",
+    onSubmit: onManualClose,
+  }
+
+  const signOffActionPluralProps = {
+    dialogTitle: "Manually close incidents",
+    dialogContentText: "Write a message describing why the incidents were manually closed",
+    question: "Are you sure you want to close these incident?",
```
podliashanyk, line 54:
```suggestion
    question: "Are you sure you want to close these incidents?",
```

podliashanyk (changes requested):
Good job so far!

Only updates for bulk re-open operation that remain (_title_, _question_)

src/components/incident/ManualCloseSignOffAction.tsx
```diff
@@ -29,37 +30,60 @@ const ManualClose: React.FC<ManualClosePropsType> = ({
...
       <ConfirmationButton
-        title={"Reopen incident"}
-        question={"Are you sure you want to reopen this incident?"}
-        onConfirm={onManualOpen}
-        ButtonComponent={ButtonComponent}
+        className={classes.dangerousButton}
         buttonProps={{
-            variant: "contained"
+            variant: "contained",
         }}
```
stveit, line 84: When i tried to but `buttonProps` inside `reopenButtonDefaultProps` I got this error:

```
Type '{ children: string; color?: "inherit" | "primary" | "secondary" | "default" | undefined; disabled?: boolean | undefined; disableElevation?: boolean | undefined; disableFocusRipple?: boolean | undefined; ... 284 more ...; buttonProps: { ...; }; }' is not assignable to type '{ title: string; question: string; ButtonComponent?: "abbr" | "address" | "article" | "aside" | "b" | "bdi" | "bdo" | "big" | "blockquote" | "button" | "caption" | "cite" | "code" | "data" | ... 44 more ... | undefined; buttonProps?: Partial<...> | undefined; }'.
 The types of 'buttonProps.variant' are incompatible between these types.
  Type 'string' is not assignable to type '"text" | "outlined" | "contained" | undefined'.  TS2322
```

So i just left it where it was

podliashanyk: buttonProps is a type defined by MUI, but no worries we don't actually need to fix this

podliashanyk (approved):
Very well!

PR #446, opened 2022-10-11
Remove useless filter list button from the table header

### LINKED ISSUE #443: Remove the "filter list"-button in the incidents-page since it currently doesn't seem to do anything

podliashanyk:
Closes #443 

Removed some redundant code in _IncidentTableToolbar_ as well.


hmpf (changes requested):
Fix the usual, then it's good to go.

hmpf (approved):

PR #457, opened 2022-11-16
Build and publish production-oriented builder Docker image as GitHub package

lunkwill42:
To implement Uninett/argus-docker#1 in a more fluent way, production-oriented Docker images of the back-end and front-end components should be built and published as packages attached to each of the source repositories.

This adds a Dockerfile to produce a production oriented image of the front-end UI application, and sets up a GitHub Actions workflow to build and publish said image as a *package* on every release made from this repo.

Because of how the site configuration is compiled in the statically generated application files, this image is just a stage 1 build image, so documentation on how to use it is added to a `README.md` file.

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -1,0 +1,43 @@
+# Much inspired from https://docs.github.com/en/actions/publishing-packages/publishing-docker-images
+name: Create and publish Argus front-end Docker image
+
+on:
+  release:
+    types: [published, edited]
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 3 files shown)


hmpf (commented):
How would we test this to see if it works? Also, does it assume that the backend is also running in Docker?

lunkwill42:
> How would we test this to see if it works? Also, does it assume that the backend is also running in Docker?

ATM, we don't. We assume that it works when our test suite is green ;-)

It doesn't assume anything about how the backend is served. Since the app runs in the user's browser, whatever backend URL was configured into the site needs to be reachable for that user's browser. Whether it is localhost, running in Docker or in AWS, doesn't matter.

Except - of course - for our slightly ridiculous problem with cookies and Feide login, where the backend needs to know the domain of the frontend. That problem needs to be solved on a different level entirely.

For the full docker compose-based environment, see https://github.com/Uninett/argus-docker

lunkwill42:
> For the full docker compose-based environment, see https://github.com/Uninett/argus-docker

We could however devise some pipeline tests for this repo, which tests the entire integrated system somehow.

hmpf (approved):

PR #466, opened 2022-12-15
Fix table loading when updating filter

podliashanyk:
Closes #460


lunkwill42 (approved):
Much better :) Now it's really visible that something is happening when I change my filters! 👍

PR #468, opened 2022-12-16
Render Federated login button only if feide login is configured

### LINKED ISSUE #463: FEIDE login button is always there
**Describe the bug**

It seems that the `LOGIN WITH FEIDE` button is always present on the Argus login page (I may be mistaken, though).

Only installations in Norway (of which there is currently only one) would be interested in that. It's entirely outside the scope of user's in any other country.

**Expected behavior**

There are two divergent paths here:

1. I assume Argus currently could be configured with any other OAuth2 identity provider than FEIDE. In that case, the word "FEIDE" in the login button is misleading.

2. If federated/OAuth2 is not explicitly configured in an Argus instance, the external login button (and the word **OR**, which appears above it) button should not be visible at all.

**Screenshots**
Screenshot 2022-12-09 at 12 27 46

**Additional context**

argus-frontend version 1.6.1

podliashanyk:
Closes #463

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -38,6 +38,7 @@ import {
   Resolver,
   ErrorCreator, 
   MetadataConfig,
+  ConfiguredLoginMethodsResponse,
 } from "./types.d";
 
 import auth from "../auth";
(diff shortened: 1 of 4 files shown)


johannaengland (commented):
Everything works as expected. 

I would wait for Uninett/Argus#562 to be merged and then include the changes that are made here.

johannaengland (changes requested):
Uninett/Argus#562 is now merged, therefore this needs to be changed to adapt to the changes in the endpoint `/login-methods/`.

johannaengland (approved):

hmpf (commented):
I'll run this one to see the button, a bit unclear from the code what it will say.

lunkwill42 (approved):
Seems fine to me. I get "OAuth2" on the extra login button in the normal case, and no button if I deconfigure the OAuth2/Dataporten/Feide backend.

👍

PR #474, opened 2023-01-30
Add global state logic and API calls for destinations

podliashanyk:
Global state for destinations is needed since destinations data will be used in both destinations settings components, and notification profiles components. 

This PR offers no functional changes, just a preparation for adding destinations to Argus

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -36,8 +36,8 @@ import {
   CursorPaginationResponse,
   ApiListener,
   Resolver,
-  ErrorCreator, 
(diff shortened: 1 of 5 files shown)


hmpf (commented):
@stveit shold check this also, espceially if nothing apperas to change when noodling about in the browser

stveit (approved):
im not all that familiar with reducers, but i dont see any obvious problems with this.
Tested this branch against master backend and it didnt seem like it causes any issues at least

Issue #465, opened 2022-12-15
Update user manual


podliashanyk:
There have been merged PRs that require updates in the MANUAL.md:

- [ ] New _Create ticket_ feature #461 
- [ ] Changed workflow for _Edit ticket_ feature #461 
- [ ] Screenshots with helper text on _bulk updates_ #444 
- [ ] Screenshots for _Severity level_ #442 
- [ ] Specify that timeslots are reusable when working with _Notification profiles_ #415

PR #475, opened 2023-01-30
Add destination settings

podliashanyk:
### Changes made:

- _Settings_ are replaced with _Destinations_

### Remaining tasks:

- [x] CSS polishes
- [x] Update changelog
- [x] Tests (separate issue #476)
- [x] Update user manual (separate issue #465)
- [x] Fix mobile view (separate issue #485)

NB! This PR must be rebased onto #474 in order for implementation to work


johannaengland (changes requested):
This looks really good! And I got it to work with the MS Teams plugin.

I only have two little comments (not sure if they belong here or if I should make a new issue for that):
1. When changing a destination (I added a label) it shows the message "Created new destination: Email: example@example.com" instead of indicating that I changed a destination. This might be confusing.
2. When adding a new destination with a label it might be nice to have the success message "Create new destination: _label of the destination_" and only using the `suggested_label` if no label is provided by the user.

stveit:
> If you edit an existing phone number, you can add up to 2 letters to the number and when you try to save you will get the error message that the number already exists, instead of it being invalid. probably a backend thing? or does a few letters just not matter for real phone numbers? seems weird

the same for creating basically, it allows you to create a phone number with 2 letters. It seems like the phone number library allows up to 2 letters but just strips them, so when editing it says "same number" because after stripping it becomes the same number

stveit:
When deleting existing destination it says " Discard Are you sure you want to discard changes?" should probably be text more fitted for deleting stuff

podliashanyk:
@stveit most of the reported problems are fixed, for others there are now PRs:

- #482 
- #483 
- #484 
- https://github.com/Uninett/Argus/issues/578

stveit:
- The fixes have made it so that there is no visual feedback when you click on something that doesnt do anything, this is good. But theres a different issue regarding what type of cursor is shown. This is prob a bad explanation, so ask me for demonstration if u need. For example, when you hover over the "SAVE" button, the cursor looks different, like you can click on what you hovering above. The same "clickable" cursor is shown if youre hovering in the white space of the 'Title' or 'Email address' box, but if you click there nothing happens. When hovering on the 'Title' or 'email address' labels, the cursor shows the "standard" cursor instead of the clickable one. I would expect the clickable cursor to be shown when hovering the title, since clicking the title will select the corresponding input box. And i would expect the standard cursor to show up when hovering over the white area where clicking has no effect
- Destination title and + button still dont look lined up to me
- When both Email and sms boxes are extended, there is no spacing between them


Tested using Firefox

stveit:
image
Also got a weird bug where the label collapses into the input box even though something is written there, but im not able to replicate it

podliashanyk:
> * The fixes have made it so that there is no visual feedback when you click on something that doesnt do anything, this is good. But theres a different issue regarding what type of cursor is shown. This is prob a bad explanation, so ask me for demonstration if u need. For example, when you hover over the "SAVE" button, the cursor looks different, like you can click on what you hovering above. The same "clickable" cursor is shown if youre hovering in the white space of the 'Title' or 'Email address' box, but if you click there nothing happens. When hovering on the 'Title' or 'email address' labels, the cursor shows the "standard" cursor instead of the clickable one. I would expect the clickable cursor to be shown when hovering the title, since clicking the title will select the corresponding input box. And i would expect the standard cursor to show up when hovering over the white area where clicking has no effect

Fixed all mentioned cursors, except for _labels_. MUI Buttons labels (which are programmatically connected to input fields) have _default_ cursors on-hover, not _pointer_ cursor as you suggest. I presume they follow the [CSS specification](https://w3c.github.io/csswg-drafts/css-ui/#cursor). I suggest we also follow this specification and not change the cursor for labels. Pointer cursor is best suited for links and buttons. 


> * When both Email and sms boxes are extended, there is no spacing between them

Fixed

podliashanyk:
> image
> Also got a weird bug where the label collapses into the input box even though something is written there, but im not able to replicate it

Could not reproduce this either 🤷‍♀️

podliashanyk:
> * Destination title and + button still dont look lined up to me

Fixed

stveit:
everything seems OK now 👍

stveit (approved):

podliashanyk:
> This looks really good! And I got it to work with the MS Teams plugin.
> 
> I only have two little comments (not sure if they belong here or if I should make a new issue for that):
> 
> 1. When changing a destination (I added a label) it shows the message "Created new destination: Email: [example@example.com](mailto:example@example.com)" instead of indicating that I changed a destination. This might be confusing.
> 2. When adding a new destination with a label it might be nice to have the success message "Create new destination: _label of the destination_" and only using the `suggested_label` if no label is provided by the user.

Fixed!

podliashanyk:
> Nearly all changes that @stveit and I requested are integrated now, the only old thing that is still open and doesn't have a separate issue is that when creating a destination there is no spacing between the "Create" and the "Discard" button.
> 
> And two more things that I noticed now:
> 
> * it is fixed that when creating a destination with a label it says "Created new destination: label", but when updating or deleting it still says "Updated/Deleted destination: suggested_label"
> * when deleting a destination it says "Are you sure you want to delete destination?" - my suggestion would be "Are you sure you want to delete the destination?" or "Are you sure you want to delete this destination?"

Fixed

PR #479, opened 2023-02-01
Add destinations to notification profiles

podliashanyk:
### Changes made in _Notification profiles:_

- Removed _Media_-selector
- Replaced _Phone number_-selector with _Destinations_-selector
- Notification profiles operate with destinations instead of phone number

### Notable details about the _Destinations_-selector:

- User can now chose **multiple** destinations per notification profile
- Destinations menu items names are either a label (user-given destination name), or a _suggested label_ (set by backend, f.e "SMS: +XXXXXXXXXX"), or an actual settings value (f.e. actual phone number, or email). But on hover event over a destination menu iten, the actual settings value (f.e. actual phone number, or email) will be displayed in a tooltip.
- Destination in the menu have checkboxes and are divided in categories after the media type that destinations belong to.

NB! The _Pluss_-button still opens a create new phone menu instead of a new destination emu. Fixing this should be a separate issue #481 after #475 is merged.

**NB! This PR is dependent on #474.**

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -16,7 +16,6 @@ import {
   PhoneNumberRequest,
   PhoneNumberSuccessResponse,
   NotificationProfilePK,
-  NotificationProfile,
   Incident,
   Event,
   EventType,
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -11,7 +11,6 @@ import {
   Filter,
   FilterPK,
   FilterString,
-  MediaAlternative,
   PhoneNumberPK,
   PhoneNumber,
   PhoneNumberRequest,
@@ -241,17 +240,17 @@ class ApiClient {
   }
 
   // NotificationProfile
-  public getNotificationProfile(timeslot: TimeslotPK): Promise<NotificationProfile> {
+  public getNotificationProfile(pk: NotificationProfilePK): Promise<NotificationProfileSuccessResponse> {
     return this.resolveOrReject(
-      this.authGet<NotificationProfile, GetNotificationProfileRequest>(`/api/v1/notificationprofile/${timeslot}/`),
+      this.authGet<NotificationProfileSuccessResponse, GetNotificationProfileRequest>(`/api/v2/notificationprofile/${pk}/`),
       defaultResolver,
       (error) => new Error(`Failed to get notification profile: ${getErrorCause(error)}`),
     );
   }
 
-  public getAllNotificationProfiles(): Promise<NotificationProfile[]> {
+  public getAllNotificationProfiles(): Promise<NotificationProfileSuccessResponse[]> {
     return this.resolveOrReject(
-      this.authGet<NotificationProfile[], GetNotificationProfileRequest>(`/api/v1/notificationprofiles/`),
+      this.authGet<NotificationProfileSuccessResponse[], GetNotificationProfileRequest>(`/api/v2/notificationprofiles/`),
       defaultResolver,
       (error) => new Error(`Failed to get notification profiles: ${getErrorCause(error)}`),
     );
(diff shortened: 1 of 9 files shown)


johannaengland (changes requested):
After creating a profile the linked destinations are all shown only by their suggested label even if they have a user-set label. I would expect to see the user-set label there as well.

podliashanyk:
> After creating a profile the linked destinations are all shown only by their suggested label even if they have a user-set label. I would expect to see the user-set label there as well.

Not sure about this one. I have testet different display values for the linked destinations, and IMHO suggested labels look most clear and exhaustive. But I will check with users after they have used the new profiles, and will probably conduct some user tests on different ways to display destinations in profiles

src/components/notificationprofile/NotificationProfileCard.test.tsx
```diff
@@ -3,39 +3,19 @@
...
 import { render, screen, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { Filter, MediaAlternative, NotificationProfileKeyed, PhoneNumber, Timeslot } from "../../api/types";
+import {
+  Destination,
+  Filter,
+  KnownProperties,
+  Media,
+  NotificationProfileKeyed,
+  PhoneNumber,
```
hmpf, line 12: Why is this still used, local validation?

src/components/notificationprofile/NotificationProfileCard.test.tsx
```diff
@@ -68,25 +48,84 @@ const filters: Filter[] = [
...
+  {
+    pk: 4,
+    label: "test_msteams",
+    media: media[2],
+    suggested_label: "MS Teams: MS TEAMS #4",
+    settings: {
+      webhook: "https://test.test"
+    }
+  },
+]
```
hmpf, line 96: There's no guarantee that the msteams plugin is installed so you can't really test it here. We might need a discussion about this.

We could have a dummy plugin that has a webhook?

src/components/notificationprofile/NotificationProfileList.test.tsx
```diff
@@ -76,13 +74,13 @@ describe("Rendering profile list with no existing timeslots", () => {
...
+            .onGet("/api/v2/notificationprofiles/")
             .reply(200, [] as NotificationProfile[])
-            .onGet("/api/v1/notificationprofiles/timeslots/")
+            .onGet("/api/v2/notificationprofiles/timeslots/")
             .reply(200, [] as Timeslot[])
-            .onGet("/api/v1/notificationprofiles/filters/")
+            .onGet("/api/v2/notificationprofiles/filters/")
             .reply(200, filters as Filter[])
-            .onGet("/api/v1/auth/phone-number/")
+            .onGet("/api/v2/auth/phone-number/")
```
hmpf, line 83: I think this endpoint will go away completely. The only place phone numbers will be stored will be in destinations.

hmpf (commented):
What's with the phone numbers everywhere? Otherwise, see comments.

podliashanyk:
@hmpf removed phone number objects from tests. But in the notification components there is is still a _Plus_ button that prompts _Add new phone number_ dialog. So I haven't removed tests for this one yet. Switching to _Add new destination_ dialog is a separate issue #481

podliashanyk:
@johannaengland fixed overlap of sticky headers :)

src/api/client.ts
```diff
@@ -36,8 +34,8 @@ import {
   CursorPaginationResponse,
   ApiListener,
   Resolver,
-  ErrorCreator, 
-  MetadataConfig,
+  ErrorCreator,
+  MetadataConfig, Media, Destination, MediaSchema, DestinationPK, DestinationRequest, NewDestination,
```
stveit, line 38: this should probably be one per line as the rest of them, looks wierd mixing it up

stveit (approved):

johannaengland (approved):
:+1:

PR #487, opened 2023-02-02
Convert logo text to paths

### LINKED ISSUE #453: Argus logo renders differently on different systems
**Describe the bug**

The full Argus logo, with the eye and the text below (primarily shown on the login screen) seems to render differently across different systems.

There are differences between different browser on my own system, but there is an even more stark difference on systems that do not have a matching font installed.

The difference is always in the "Argus" text itself, and is likely connected to two issues:

1. How SVG text is rendered by different rendering engines.
2. Which fonts are installed on the system that renders the logo (the logo specifies the font *Futura-Bold*)

While we *do* want to keep the source SVG file with the text intact, we should also produce a version of the SVG file where the text has been converted to shapes. This is the only way to ensure consistent rendering between systems.

**To Reproduce**

Browse the Argus login page on different combinations of operating systems and browsers.

**Expected behavior**

The logo should be more or less identical across all systems.

**Screenshots**

This is from Firefox 105 on MacOS 12.6:
Screenshot 2022-10-28 at 11 36 11

This is from Safari 16.0 on MacOS 12.6:
Screenshot 2022-10-28 at 11 37 26

This is from Microsoft Edge 106.0.1370.37 on Windows 10:
Screenshot 2022-10-28 at 12 34 42

lunkwill42:
The SVG logo files include the "Argus" text with a font specifications. Unfortunately, different systems will render the text differently, depending both on the installed fonts and the rendering engine used.

This renames the old SVG logo files to keep the originals, and replaces the logos used in the web interface with ones where the text has been rendered to SVG paths instead. This ensures more consistent rendering of the SVGs across systems.

Fixes #453


podliashanyk (approved):
👍🏻

hmpf (commented):
I don't quite understand what I'm supposed to look for here. In both Firefox and Vivaldi, logo_dark.svg (deleted) looks much better than logo_dark.svg (added).

logo_white.svg (deleted) has too much space between A and R, and logo_white.svg (added) is eye-pokingly horrible.

hmpf (commented):
image

I used to noodle with typography as a hobby... this would be called potato-graphics if it were a game.

image

For white, "added" actually looks better than "deleted".

stveit:
Ive centered the G on the eye using inkscape aligning tool, so its centered on the "hitbox" of the G to the "hitbox" of the eye. With this font the apex of the G curve is a bit to the right, so this part is not centered. it looks a bit weird to me but i will defer to the exper @hmpf

hmpf (approved):
Looks really good!

lunkwill42 (commented):
While this looks much closer to the original, I can't help but ask if the font used for rendering is entirely the same?

Attached is the original logo artist's PNG rendering of the logo, for comparison:

no frame-8

stveit (approved):

lunkwill42:
Conclusion: Fonts are nuts, and we're not designers 😆

PR #488, opened 2023-02-02
Add description to acknowledment body

podliashanyk:
Bugfix

description inside EventBody is for api v1 and description outside EventBody is for api v2

### DIFF
--- a/src/components/incident/CreateAckSignOffAction.tsx
+++ b/src/components/incident/CreateAckSignOffAction.tsx
@@ -29,6 +29,7 @@ const CreateAck: React.FC<CreateAckPropsType> = ({ onSubmitAck, signOffActionPro
       },
       expiration: selectedDate && selectedDate.toISOString(),
       timestamp: new Date().toISOString(),
+      description: msg,
     });
     setSelectedDate(null);
   };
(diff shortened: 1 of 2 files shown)


src/components/incident/CreateAckSignOffAction.tsx
```diff
@@ -29,6 +29,7 @@ const CreateAck: React.FC<CreateAckPropsType> = ({ onSubmitAck, signOffActionPro
       },
       expiration: selectedDate && selectedDate.toISOString(),
       timestamp: new Date().toISOString(),
+      description: msg,
```
johannaengland, line 32: Do I see that right that you now post the description twice? Once within event and once outside of it? I think it that case it would be better to have to different bodies: one for posting to v1 and one for posting to v2

johannaengland (changes requested):
See inline comment.

src/components/incident/CreateAckSignOffAction.tsx
hmpf: You will likely have two different timestamps as well.

johannaengland (approved):

hmpf (approved):
Tested manually, works.

PR #496, opened 2023-03-21
Improve error handling in Login view

### LINKED ISSUE #451: Misleading error message on login screen when frontend cannot connect to backend
When trying to log in while there is no backend to connect to this is shown:

image

An error message indicating that it couldnt connect to Argus or something like that might be better than to make the user 
think that their credentials are wrong.

podliashanyk:
### Implemented changes:
There are now 3 different ways to inform user about errors in _Login view:_
1. Error appbar in case of netwrok/server problems
2. Helper text informing about wrong credentials if userpass request failes
3. Error snackbar in case get user request fails

Closes #451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ This file documents all changes to Argus-frontend. This file is primarily meant 
 
 ## [Unreleased]
 
+### Added
+- Handling of both network errors, and errors with fetching existing user in _Login view_ .
+
 
 ### Fixed
 - Bug with misleading error message in Login view on connection problems.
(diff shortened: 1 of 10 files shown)


lunkwill42 (approved):
This works much better than before :)

But when recently discussed CORS problem occurs, I still get this message, which is somewhat imprecise:

> Problems connecting to server... Check your connection.

podliashanyk:
> But when recently discussed CORS problem occurs, I still get this message, which is somewhat imprecise:
> 
> > Problems connecting to server... Check your connection.

I have now created a separate issue for this #498 and suggested two alternatives on giving users the feedback in case of CORS errors. Please let me know if you have any thoughts on this issue.

Merging this PR :)

PR #505, opened 2023-04-18
Upgrade react-scripts to 5.0.1

podliashanyk:


hmpf (changes requested):
Curious, why so many lines with eslint-disable-next-line?

I can't get this to work against newest backend master, exception from backend when clicking on "Profiles".

Otherwise most of the changes seems to be turning off eslint.

hmpf (approved):
Survived quick manual test

PR #506, opened 2023-04-21
Remove all use of filter_string

podliashanyk:
Can be merged when Uninett/Argus#619 is completed


hmpf (commented):
This looks good, but I will not approve until I have a version of the backend that I can test against.

johannaengland (approved):

johannaengland (approved):
Tested against the newest backend master and everything works as it should

PR #507, opened 2023-04-21
Upgrade versions of GitHub Actions

### LINKED ISSUE #495: Upgrade docker images workflow to avoid deprecated GitHub Actions features
**Describe the bug**

The docker images workflow uses some actions that use deprecated versions of Node.js. I.e. the workflow runs are full of message like this:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9, docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38, docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

These workflows should be updated:

https://github.com/Uninett/Argus-frontend/blob/df2239b30761c54f62f8aaa7d56d0730e6357e1a/.github/workflows/docker-images.yml#L2

podliashanyk:
Closes #495


.github/workflows/docker-images.yml
```diff
@@ -21,22 +21,22 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Log in to the Container registry
-        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        uses: docker/login-action@v2
```
hmpf, line 24: Version numbers feel so much safer :)

hmpf (changes requested):
Save a newline at the end of node.js.yml to shut up the differ.

Is there a pre-commit for this set up yet?

hmpf (approved):

PR #517, opened 2023-04-27
User manual: Polish links and references

podliashanyk:
Closes #509


hmpf (approved):
Maybe we should have a link-checker for the manual as part of the tests? For manual-internal links.

Issue #494, opened 2023-03-17
Polish: Make the lines under the fields on the "Notification profiles"-page line up


hmpf:
I've only tested this on Vivaldi, which uses the same renderer as chrome.

_Originally noted by @hmpf in https://github.com/Uninett/Argus-frontend/issues/492

podliashanyk:
The field that isn't lined up is the _Filters-selector_. It is rendered as MUI's `Autocomplete`. All other fields are rendered as MUI's `Select`.

PR #529, opened 2023-09-20
Add name to notification profiles

### LINKED ISSUE #524: Notification profiles needs their own visible names in frontend
When timeslots could be used only by one notification profile, they shared a name. Now that a timeslot can be used several times, it is still only the timeslot names that are available in the frontend. A profile/name can be faked by combining timeslot name + filter name, but still it would be nice to be able to call a profile something like "testing", regardless of timeslots and filters involved.

podliashanyk:
Closes #524, #494

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,14 @@ This file documents all changes to Argus-frontend. This file is primarily meant 
 
 ## [Unreleased]
 
+
 ### Added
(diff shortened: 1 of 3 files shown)


johannaengland (approved):
All looks very good! Very nice two-in-one issue fix!

One thing that confused me a bit is that the placeholder for the name is "Profile Name", but the label is "Title". I would expect the label to be "Name", but that might also just be personal preference - you are the expert here!

PR #525, opened 2023-06-27
Fix destinations fetch in notification profiles

podliashanyk:
Closes #523

PR #530, opened 2023-09-21
Fix bug with endless calls to destinations endpoint from the profiles page

podliashanyk:
Bug was introduced in #525


johannaengland (approved):
At least this is very easy to check :D 
Perfectly fixes it!

lunkwill42 (approved):
Tested and found ok :+1:

PR #536, opened 2024-01-30
Make it possible to change configuration variables at runtime

### LINKED ISSUE #456: Research alternative ways to configure the frontend application
The frontend application configuration is defined in `config.tsx` - and the values are provided as environment variables at *build time*.

If we have a goal to provide a Docker image with the Argus frontend in it, this still needs to build everything from scratch at deployment time, since all the configuration is statically compiled into the resulting JavaScript files. This makes every deployment situation "fat" and time-consuming.

I think it's desirable for frontend deployments to be as lightweight as possible. Can we find ways to generate a static file with JavaScript or JSON-based configuration from the run-time environment, without involving the entire npm build framework, and have the React application load this file at runtime?

Surely this situation is not unique in the world - so it's likely someone else has made useful solutions we can learn fro.

podliashanyk:
Fixes #456?

The implementation here is heavily inspired by the global config option described here https://profinit.eu/en/blog/build-once-deploy-many-in-react-dynamic-configuration-properties/

### Implemented changes:

- [x] Configuration can now be changes at runtime in dev and prod environments via variables provided in `runtime-config.json` file
- [x] Updated CHANGELOG
- [x] Updated docs on frontend

### Expected behaviour:

- Default config values are **always** used in _test env_.
- Default config values are used in _dev env_ unless `/public/runtime-config.json` is added. Updating `/public/runtime-config.json` at runtime will immediately trigger changes to the application when running the application in dev via `docker-compose up`.
- There is no default config in _prod env_. `%FRONTEND_BASE_URL%/runtime-config.json` file **must** be served or else the app will not run.

### Following is to be expected when `runtime-config.json` is provided:

- `backendUrl` and `cookieDomain` are required, otherwise the app will fail.
- if `frontendVersion` or `apiVersion` values are provided, they are silently ignored.
- if any other unknown variables are provided, they are silently ignored.

### Further work:
- document new configuration process on readthedocs (https://github.com/Uninett/Argus/issues/731)
- display error message in the view if configuration fails (per now on error the page is blank and error is printed to console) (#542)
- add typecheks for config values at runtime (#543)

### DIFF
--- a/docker/runtime-config-template.json
+++ b/docker/runtime-config-template.json
@@ -4,17 +4,14 @@
     "enableWebsocketSupport": ${REACT_APP_ENABLE_WEBSOCKETS_SUPPORT},
     "useSecureCookie": ${REACT_APP_USE_SECURE_COOKIE},
     "debug": ${REACT_APP_DEBUG},
-    "cookieDomain": "localhost",
-
+    "cookieDomain": "${REACT_APP_COOKIE_DOMAIN}",
     "defaultAutoRefreshInterval": ${REACT_APP_DEFAULT_AUTO_REFRESH_INTERVAL},
     "realtimeServiceMaxRetries": ${REACT_APP_REALTIME_SERVICE_MAX_RETRIES},
-
     "use24hTime": true,
     "timestampDateFormat": "yyyy-MM-dd",
     "timestampTimeFormat": "HH:mm:ss",
     "timestampTimeNoSeconds": "HH:mm",
     "timestampTimezoneOffsetFormat": "xxx",
     "timestampFormat": "{date} {time}{timezone_offset}",
-
     "showSeverityLevels": true
 }


lunkwill42:
### An attempt to summarize an offline discussion:

This doesn't seem to translate into an actual runtime configuration file.  The values of `runtime-config.json` are built into the JS bundles at compile time (as the `require` directive is a compile-time directive), so this suggested solution seems to be essentially identicalto the current solution (i.e. it fetches build-time config from a file rather than from the environment, but it is still build-time configuration).

What we want is an actual **separate** file with *only* configuration variables, which is loaded from the web server as the React application starts.  This file should be editable during runtime, using any editor, like e.g. *vim*, and the changes should take effect merely by hard-reloading the application in your browser.

When deployed as a container, said configuration file should be generated from the container's environment variables as the container starts.  An example of a current production-oriented deployment can be found in https://github.com/Uninett/argus-docker/blob/master/frontend/Dockerfile . This is a two-stage `Dockerfile`.  The first stage uses NPM to build the Argus frontend, having the configuration variables come from build-time arguments.  The second stage is a simple nginx-based container, where the static files to be served are copied from the first container (so we have a resulting minimal image without NPM, just with a web server and some static files).  Ideally, the config variables that come as build `ARG`s in the first stage should rather come as `ENV` vars in the second stage.  The run command then needs to be a two-stage script that first generates the static config file from these environment variables (maybe using a simple templating command like [envsubst](https://man7.org/linux/man-pages/man1/envsubst.1.html), and then launches nginx as before.

lunkwill42 (commented):
I have tested this with argus-docker, and in our PaaS environment by updating the K8s deployment definitions for a review site, and things seem to be working smoothly - and everything has fewer deploy steps now that the same docker image can be used for all deployments.

Updating docs and ensuring there is good error feedback if the runtime-config is missing or broken is all that is left to do, IMO.

lunkwill42:
I've added Uninett/argus-docker#3 to track this when it hits a new release.

README.md
```diff
@@ -85,9 +85,16 @@ Note that the website will automatically reload as you edit the code.
...
 
-Configuration options for the Argus frontend are located in `src/config.tsx`. All of these options can be set using environment variables when running the frontend under the Node server (or in Docker Compose). However, for production deployment, you would normally build the application and serve all the resulting static files using a regular web server, like Apache or Nginx; in this case, `config.tsx` cannot read server environment varibles, and should be hard coded instead.
+Default configuration options for the Argus frontend are located in `src/config.tsx`. All of these options can be set using **environment** variables when running the frontend under the Node server (or in Docker Compose), or by providing **configuration** variables in `runtime-config.json` file.
 
-These environment variables are available:
+#### Development environment
+Either provide **environment** variables when running the frontend under the Node server (or in Docker Compose), or add `runtime-config.json` file with the **configuration** variables to the `/public` folder.
+
+#### Production environment
+Serve `./runtime-config.json` file with the **configuration** variables.
```
lunkwill42, line 94: How about an example configuration file?  I don't find it quite intuitive what this file needs to look like.  You refer to the file both as `./runtime-config.json` *and* also mention a `/public` folder - but what is the format of the file, and on which URI should it be served?

README.md
```diff
@@ -85,9 +85,16 @@ Note that the website will automatically reload as you edit the code.
...
 
-These environment variables are available:
+#### Development environment
+Either provide **environment** variables when running the frontend under the Node server (or in Docker Compose), or add `runtime-config.json` file with the **configuration** variables to the `/public` folder.
+
+#### Production environment
+Serve `./runtime-config.json` file with the **configuration** variables. 
+
+#### Variables
+These **environment** variables are available:
```
lunkwill42, line 97: You should also mention somewhere that these environment variables can also be used as described here if using the production-oriented Docker image defined in `./docker/Dockerfile`.  This image will produce a `runtime-config.json` from the environment variables exported to the container when it starts.

Perhaps even link to the `docker/README.md` file (which I see I also need to update, since it is outdated with the changes I contributed to this PR.

lunkwill42 (changes requested):

docker/runtime-config-template.json
```diff
@@ -1,20 +1,17 @@
...
-    "debug": ${REACT_APP_DEBUG},
-    "cookieDomain": "localhost",
-
-    "defaultAutoRefreshInterval": ${REACT_APP_DEFAULT_AUTO_REFRESH_INTERVAL},
-    "realtimeServiceMaxRetries": ${REACT_APP_REALTIME_SERVICE_MAX_RETRIES},
-
+    "enableWebsocketSupport": "${REACT_APP_ENABLE_WEBSOCKETS_SUPPORT}",
+    "useSecureCookie": "${REACT_APP_USE_SECURE_COOKIE}",
+    "debug": "${REACT_APP_DEBUG}",
+    "cookieDomain": "${REACT_APP_USE_SECURE_COOKIE}",
```
lunkwill42, line 7: Question: These two variables are supposed to be booleans, but this forces them to JSON strings.  I see this is handled if reading directly from the environment under npm, but how is this handled when reading `runtime-config.json` directly?

lunkwill42: Sorry, I did not intend to flag `cookieDomain` here, but rather `enableWebsocketSupport` (in addition to `debug`)

PR #553, opened 2024-04-10
Add link to admin page

### LINKED ISSUE #537: Add a link to the admin page
Uninett/Argus#712 introduces that in API v2 when accessing the endpoint `/api/v2/auth/user/` we have a field called `admin_url` that is the link to the Django admin, where user administration and other things can be done. 

@lunkwill42 and I discussed that it would be best to add an item with the label "Admin" here:
Screenshot from 2024-02-02 11-06-39

`admin_url` can be empty (if the user is not staff), so it should only be shown if it is not empty.

podliashanyk:
Closes #537

### DIFF
--- a/src/components/header/Header.tsx
+++ b/src/components/header/Header.tsx
@@ -234,6 +234,7 @@ const Header: React.FC<HeaderPropsType> = () => {
 
       {user.admin_url &&
         <MenuItem className={classNames(style.menuItemWithIcon)}
+                  aria-label="Link to site administartion"
                   component="a"
                   href={user.admin_url}
                   target='_blank'


johannaengland (changes requested):
I think it looks great, I really like the new design with the icons and the horizontal divider.

My only problem is the fact that the `admin_url` is an absolute url and I think you're treating it as a relative url, which results in if I get `admin_url: http://localhost:8000/admin/` then the link I'm clicking on is http://localhost:8080/http://localhost:8000/admin/, which interestingly enough first gave me a 404 error, but now actually leads me to the admin.
But it still should only link to the admin_url exactly how it's received.

johannaengland (approved):
Perfect! Tested both with an admin account and a normal account and all works as expected!

lunkwill42 (approved):
Nice!

PR #557, opened 2024-04-15
Add source internal id detail

podliashanyk:
Closes #499 

Changes:

- [x] Display source internal ID
- [x] Write tests
- [x] Changelogs


lunkwill42 (approved):
Looks fine to me, but it reminds me again that we really should look redesign the details dialog...

image

johannaengland (approved):

PR #560, opened 2024-04-19
Show spinner while ticket is being generated

### LINKED ISSUE #486: Add visual feedback while generate ticket event is loading
Per now there is no visual feedback only when event is completed, the ticket is opened in a new Tab

podliashanyk:
Closes #486

### DIFF
--- a/changelog.d/560.added.md
+++ b/changelog.d/560.added.md
@@ -1,0 +1,1 @@
+Visual feedback via spinner in the _Create ticket_-button that ticket is being automatically generated.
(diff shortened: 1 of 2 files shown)


johannaengland (approved):
Works as expected, the only small thing I noticed is that when I have Zoom at 100% the button moves when the loading icon is shown:
1. Before generating a ticket:
Screenshot from 2024-04-22 14-51-43
2. While the ticket is generated:
Screenshot from 2024-04-22 14-47-53

3. In comparison at 90% Zoom while the ticket is generated:
Screenshot from 2024-04-22 14-48-28

podliashanyk:
This is already fixed in the new frontend

PR #563, opened 2024-04-22
Copy unreleased notes to changelog

podliashanyk:
In some PRs changelog entries were forgotten to be updated (only notes were updated instead). This PR makes chengelog more up to date.
As mentioned in #559:

> Before it is done, some entries need to be manually moved from `NOTES` to `CHANGELOG`. This is because in the following PRs for some reason I have only updated the `NOTES` file (which further proves that keeping both files can easily lead to confusion when contributing):
> 
> * [Add source internal id detail #557](https://github.com/Uninett/Argus-frontend/pull/557)
> * [Add missing article in in ticket generation modal #552](https://github.com/Uninett/Argus-frontend/pull/552)
> * [Add link to admin page #553](https://github.com/Uninett/Argus-frontend/pull/553)

### DIFF
--- a/changelog.d/557.changed.md
+++ b/changelog.d/557.changed.md
@@ -1,0 +1,1 @@
+_Incident ID in \<its source system\>_ field to incident details view, which displays ID by which incident is referenced internally in its source system.


changelog.d/557.changed.md
```diff
@@ -0,0 +1 @@
+_Incident ID in \<its source system\>_ field to incident details view, which displays ID by which incident is referenced internally in its source system.
```
hmpf, line 1: I don't understand what this is saying but it's text and we're in a hurry so ok

hmpf (approved):

PR #564, opened 2024-04-24
Add short howto for release process

hmpf:


RELEASE_HOWTO.md
```diff
@@ -0,0 +1,21 @@
...
+4.  Update the version number in the file `package.json` first, easy to forget.
+    It's close to the top of the file.
+5.  Run `npm install` to upgrade the `package-lock.json`-file
+6.  Update the changelog with `towncrier build --version yournewversionnumber`.
+    Feel free to massage the generated text!
+7.  Commit the changes from steps 4 to 6 in the same commit.
+8.  Tag the commit made in step 7 with an annotated tag with the new version
+    number. "Annotated" means it needs the `-a` flag, and a message. You can
+    give the message in the same go with the `-m`-flag. Sum up the most
+    important thing in the release in the message.
```
johannaengland, line 18: Here I would love an example, because from the text to me it would not be clear how to use the `-a` or `-m` flags

johannaengland (changes requested):

RELEASE_HOWTO.md
hmpf: I won't be adding any more examples, because I'm the one doing the releases right now and I need just hints. Great detail can be added later, in other PRs.

RELEASE_HOWTO.md
```diff
@@ -13,9 +13,14 @@
     Feel free to massage the generated text!
 7.  Commit the changes from steps 4 to 6 in the same commit.
 8.  Tag the commit made in step 7 with an annotated tag with the new version
-    number. "Annotated" means it needs the `-a` flag, and a message. You can
-    give the message in the same go with the `-m`-flag. Sum up the most
-    important thing in the release in the message.
+    number.
+
+    ```
+    git tag -a -m "Mainly bugfixes" v1.5.1
```
johannaengland, line 19: Thank you for that! 

Just curious in the Argus documentation we have 
```
git tag -m 'Post release bugfixes' v1.5.1
```
as the example. Is that lacking the `-a` flag?

johannaengland (approved):

RELEASE_HOWTO.md
lunkwill42: `-a` is implied when `-m` is used without other options to add signatures to the tag, so the examples are effectively the same.

lunkwill42 (approved):

hmpf:
Project is mothballed.

PR #575, opened 2024-08-27
Remove .gitkeep from ignored towncrier files

johannaengland:
Since v24.8.0 gitkeep is automatically ignored


johannaengland:
Project is mothballed, closing.

PR #586, opened 2024-10-15
Bump python and postgres versions when running Argus in docker

### LINKED ISSUE #587: Update python (to 3.10) in docker

podliashanyk:
Using python 3.8 leads to `ERROR: Package 'argus-server' requires a different Python: 3.8.20 not in '>=3.9'`.

Closes #587, #588


hmpf (commented):
We run on 3.10 i production so I don't see why we shouldn't do that here as well

hmpf:
Worth doing.

lunkwill42 (approved):

